PR #39, opened 2015-01-06
Fix ellipsis value in app

mmilano:


prakash-alpine:
Michael: This looks good to me.

PR #44, opened 2015-01-28
Update version.rb

nkapinos:
Updating to version 5.2


jthompson6:
LGTM, since it is the same as Jason's commit d84f47c.

PR #48, opened 2015-02-21
Release 5.3 LDAP Authentication Improvements

prakash-alpine:
This pull request contains changes to support flexible LDAP schema and configuration requested for authenticating users using group membership by Fidelity, BMS and other customers. 
JIRA issues: DEV-6135, DEV-6178, DEV-8530


app/services/ldap_client.rb
```diff
@@ -21,32 +29,241 @@ def search(username)
...
+  end
+
+  # Fetch list of users from a specified LDAP group and add them to Chorus users table. In order to authenticate with LDAP
+  # we are required to create those users in Chorus database.
+
+  def add_users_to_chorus(groupname)
+
+    license = License.instance
+    dev_licenses = license[:developers]
+    dev_licenses = 2
```
AndrewLngdn: looks like line 153 should be taken out?

prakash-alpine: Good catch. This was one of the missed commits. Checked in the fix.

app/services/ldap_client.rb
```diff
@@ -21,32 +29,239 @@ def search(username)
...
+      end
+
+      user_entries.first
+    end
+  end
+
+  # Fetch list of users from a specified LDAP group and add them to Chorus users table. In order to authenticate with LDAP
+  # we are required to create those users in Chorus database.
+
+  def add_users_to_chorus(groupname)
```
AndrewLngdn: groupname is never used in this function. I think we just take the group from ldap.properties config

PR #67, opened 2015-06-22
Fix DEV-6748 - Gray 'Create Chart' button style when disabled.

curran:
Added CSS that displays 'Create Chart' button text as gray if the button is disabled.


prakash-alpine:
Curran: Is button still clickable? Is should not be.

curran:
No, the button is not clickable.

It was already not clickable before, but was displaying like a clickable button.

prakash-alpine:
LGTM

PR #68, opened 2015-06-24
DEV-11140 

atul-alpine:
fixed toast message for Job create and update


atul-alpine:
I had created this pull request in comparison with Chorus:Release_5.4 and not with master hence closing it. Created another one which compares with master

PR #71, opened 2015-06-30
Added VisEngine and Chiasm code into Chorus using Git Submodules

curran:
This pull request is the first step of getting the VisEngine and Chaism code into the Chorus directory tree. The main purpose is to make sure this does not break the production build. The changes include:

 * `components/visEngine` was added as a Git submodule.
 * The Chorus Gemfile was modified to depend on `components/visEngine`
 * The Chiasm bundle is included in the Rails asset pipeline (the bundle excludes D3 and Lodash, and expects them to be loaded already - the ordering for this is determined by `app/assets/javascripts/application.js`)

In order for this to work, `git submodule init` and `git submodule update` must be executed before starting 
Chorus. If everything works as expected, you should have `Chiasm` available as a global variable in the 
JavaScript console, which looks like this:

Curran:  the git commands mentioned above needs to be included in the chorus startup script. We can't expect everyone to run this commands before starting Chorus. Also, we need to discuss with Nate how to package this into Chorus installer.

```


> Chiasm
application.js:121324 function (container){
  var chiasm = Chiasm(container);
  chiasm.plugins.layout = layout;
  chiasm.plugins.barChart = barChart;
  chiasm.plugins.lineChart = lineChart;
  chiasm.plugins.scatterPlot = scatterPlot;
...
```


curran:
Also, to test that visEngine is being mounted, you can access

`http://localhost:3000/vis_engine/reduce_data`

and you should see

`[{"x":2},{"x":3},{"x":4}]`

curran:
Here's an update as to the status of the Chiasm integration and what I have been up to:

* Merged Chorus branches "visualizations" (which contains the UI integration) and "chiasm-visEngine" (which contains the updated bundle that excludes Underscore and D3).
* Rebased the merged branches on the current master branch
* Ran the database migration script for the Roles and Permssions updates
* Modified the Chorus UI integration to use the newer Chiasm bundle, which involves changing from using RequireJS to load the module to simply accessing the global variable exposed by the bundle
* Chiasm was developed using Lodash, and was using a function _.cloneDeep, which is not available in Underscore. I modified Chiasm to no longer use this function, so it works with the version of Underscore present in Chorus

The most up-to-date Chaism integration lives in this chiasm-visEngine branch, and has a working Bar Chart that connects with the legacy API that computes the aggregation via SQL (which does not work on Hadoop data sets).

Steps to take next include:

* Modify the visEngine REST API to return a fixed data table (as a temporary measure in preparation for the random sampling API)
* Modify the client to compute the aggregations using the [data-reduction module](https://github.com/curran/data-reduction)
* Make this integration work with Line Chart

prakash-alpine:
@curran 
The git commands mentioned above (git submodule init, git submodule update)  needs to be included in the chorus startup script. We can't expect everyone to run this commands before starting Chorus. Also, we need to discuss with @nkapinos  how to package this into Chorus installer.

curran:
Refactored the bundling logic for the Chiasm integration, so the custom VisEngine Chiasm plugins can be authored using Node.js modules. Here's [where the bundle lives](https://github.com/curran/visEngine/tree/4536eb7dd2fda29982b368ea2ad363a217686096/app/assets/javascripts/vis_engine), and here's the [Chorus page that uses it](https://github.com/Chorus/chorus/blob/f0bfabbd5b6ea82d466dce3247057a7ab57e1c50/app/assets/javascripts/views/dataset/dataset_content_details.js#L294).

app/assets/javascripts/application.js
```diff
@@ -39,6 +39,8 @@
 //= require_tree ./vendor
 //= require messenger
 
+//= require vis_engine/application
```
kevin-alpine: Hi Curran,

Why do you include this file twice?  It is some strangeness with load order?

KT

alpinebuilder:

app/assets/javascripts/application.js
curran: This is just a mistake, it should only be there once.

kevinmtrowbridge:
We got Chiasm in in 5.8!! But, not via Git submodules, instead by Components.

Thanks so much for your persistence on getting Chiasm into Chorus, @curran!

PR #78, opened 2015-07-14
DEV-11140: Show job name in "toast" (message popup) when creating / editing

kevin-alpine:
This is a smallish change: 
* a few keys in the Messages_en.properties file
* A tweak to a backbone file ...
See original ticket here: https://alpine.atlassian.net/browse/DEV-11140

PR #77, opened 2015-07-14
DEV-11635 Rename "Help" button to "Documentation"

kevin-alpine:
A very small change.  See original ticket here: https://alpine.atlassian.net/browse/DEV-11635

PR #79, opened 2015-07-14
DEV-11396: Display job creation/deletion in workspace activity stream

kevin-alpine:
This is a medium feature, it touches the entire stack, but it is very vertically specialized.

It's very similar to #80.

See original ticket here: https://alpine.atlassian.net/browse/DEV-11396

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -18,6 +18,8 @@ class Job < ActiveRecord::Base
   has_many :job_results, :order => :finished_at
   has_many :activities, :as => :entity
   has_many :events, :through => :activities
+  has_many :comments, :through => :events
+  has_many :most_recent_comments, :through => :events, :source => :comments, :class_name => "Comment", :order => "id DESC", :limit => 1
   has_many :job_subscriptions
   has_many :success_recipients, :through => :job_subscriptions, :source => :user, :conditions => ['job_subscriptions.condition = ?', 'success']
   has_many :failure_recipients, :through => :job_subscriptions, :source => :user, :conditions => ['job_subscriptions.condition = ?', 'failure']
@@ -35,6 +37,9 @@ class Job < ActiveRecord::Base
 
   scope :ready_to_run, -> { where(enabled: true).where(status: IDLE).where('next_run <= ?', Time.current).order(:next_run) }
   scope :awaiting_stop, -> { where(status: STOPPING).where('updated_at < ?', 1.minutes.ago) }
+
+  after_create :create_job_created_event , :if => :current_user
+  after_destroy :create_job_deleted_event , :if => :current_user
 
   def self.eager_load_associations
     [
@@ -210,4 +215,19 @@ class Job < ActiveRecord::Base
   def owner_can_edit
     errors.add(:owner, :JOB_OWNER_MEMBERSHIP_REQUIRED) unless (owner.admin? || workspace.members.include?(owner))
   end
+
+  def create_job_created_event
+    Events::JobCreated.by(current_user).add(
+        :job => self,
+        :workspace => workspace,
+    )
+  end
+
+  def create_job_deleted_event
+    Events::JobDeleted.by(current_user).add(
+        :job => self,
+        :workspace => workspace,
+    )
+  end
+
 end
--- a/public/messages/Messages_en.properties
+++ b/public/messages/Messages_en.properties
@@ -278,6 +278,13 @@ activity.header.WorkspaceChangeName.default={{actorLink}} updated the name of wo
 activity.header.TableauWorkbookPublished.default={{actorLink}} published {{datasetType}} {{datasetLink}} to the Tableau workbook {{tableauWorkbookLink}} in project {{tableauProjectLink}}
 activity.header.TableauWorkfileCreated.default={{actorLink}} added file {{workfileLink}} based on the {{datasetType}} {{datasetLink}} to workspace {{workspaceLink}}
 
+activity.header.JobCreated.default={{actorLink}} created job {{jobLink}} in workspace {{workspaceLink}}
+activity.header.JobCreated.without_workspace={{actorLink}} created job {{jobLink}}
+
+activity.header.JobDeleted.default={{actorLink}} deleted job {{jobLink}} in workspace {{workspaceLink}}
+activity.header.JobDeleted.without_workspace={{actorLink}} deleted job {{jobLink}}
+
+
 activity.insight=Insight
 activity.more=More
 activity.note=Note
@@ -497,7 +504,7 @@ dataset.associate_multiple.toast=Associated {{count}} datasets with the workspac
 dataset.associate.search=Find workspaces
 dataset.associate.empty.placeholder=No workspaces found
 dataset.attach=Attach Datasets
-dataset.browse_more_workspace=There is not any data in this workspace. Search or {{linkText}} to find and associate data so that it is available to the workspace. 
+dataset.browse_more_workspace=There is not any data in this workspace. Search or {{linkText}} to find and associate data so that it is available to the workspace.
 dataset.browse.linkText=browse the available data sources
 
 dataset.browse_more_data_source=You do not have datasets in this schema.
@@ -1263,9 +1270,9 @@ hdfs_connection_parameters.dialog.remove_pair=Delete
 hdfs_connection_parameters.dialog.save=Save
 
 invalid_route.title=Sorry, that resource could not be found.
-invalid_route.text=We could not find what you were trying to access. It might not exist anymore, or there may be a problem with the link. 
+invalid_route.text=We could not find what you were trying to access. It might not exist anymore, or there may be a problem with the link.
 invalid_route.dataset.title=Sorry, the table/view could not be found.
-invalid_route.dataset.content=The table/view may be deleted, is being created, or you do not have permissions to view it. 
+invalid_route.dataset.content=The table/view may be deleted, is being created, or you do not have permissions to view it.
 invalid_route.database_index.title=Sorry, the data source is not reachable.
 invalid_route.database_index.content=The data source may be offline or you do not have sufficient permissions to view it.
 invalid_route.hadoop_dataset.title=Sorry, the dataset contents could not be retrieved.
--- a/spec/javascripts/presenters/activity_presenters_spec.js
+++ b/spec/javascripts/presenters/activity_presenters_spec.js
@@ -2180,7 +2180,7 @@ describe("chorus.presenters.Activity", function() {
             it("has the right header html", function() {
                 this.presenter.options.displayStyle = ["without_workspace"];
                 expect(this.presenter.headerHtml().toString()).toMatchTranslation(
-                    "activity.header.JobCreated.without_workspace", this.translation_params
+                    "activity.header.JobDeleted.without_workspace", this.translation_params
                 );
             });
         });
@@ -2189,7 +2189,7 @@ describe("chorus.presenters.Activity", function() {
             it("has the right header html", function() {
                 this.presenter.options.displayStyle = ["default"];
                 expect(this.presenter.headerHtml().toString()).toMatchTranslation(
-                    "activity.header.JobCreated.default", this.translation_params
+                    "activity.header.JobDeleted.default", this.translation_params
                 );
             });
         });
--- a/spec/javascripts/presenters/activity_presenters_spec.js
+++ b/spec/javascripts/presenters/activity_presenters_spec.js
@@ -2125,6 +2125,77 @@ describe("chorus.presenters.Activity", function() {
         });
     });
 
+    context("job created event", function () {
+        beforeEach(function () {
+            this.model = backboneFixtures.activity.jobCreated();
+            this.presenter = new chorus.presenters.Activity(this.model);
+            this.actor = this.model.actor();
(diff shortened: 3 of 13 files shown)


kevin-alpine:
This feature will take a bit to review, and it's getting late.  I will review this first thing tomorrow morning, thanks for your patience, @prakash-alpine and @atul-alpine!

app/models/job.rb
```diff
@@ -36,6 +38,9 @@ class Job < ActiveRecord::Base
   scope :ready_to_run, -> { where(enabled: true).where(status: IDLE).where('next_run <= ?', Time.current).order(:next_run) }
   scope :awaiting_stop, -> { where(status: STOPPING).where('updated_at < ?', 1.minutes.ago) }
 
+  after_create :create_job_created_event , :if => :current_user
+  after_destroy :create_job_deleted_event , :if => :current_user
```
kevin-alpine: Why `:if => :current_user`?

spec/javascripts/presenters/activity_presenters_spec.js
```diff
@@ -2125,6 +2125,77 @@ describe("chorus.presenters.Activity", function() {
...
+                    "activity.header.JobCreated.without_workspace", this.translation_params
+                );
+            });
+        });
+
+        context("with workspace", function() {
+            it("has the right header html", function() {
+                this.presenter.options.displayStyle = ["default"];
+                expect(this.presenter.headerHtml().toString()).toMatchTranslation(
+                    "activity.header.JobCreated.default", this.translation_params
```
kevin-alpine: I think this should be `activity.header.JobDeleted.default` ...

spec/javascripts/presenters/activity_presenters_spec.js
```diff
@@ -2125,6 +2125,77 @@ describe("chorus.presenters.Activity", function() {
...
+                workspaceLink: linkTo(this.workspace.showUrl(), this.workspace.name()),
+                jobLink: linkTo(this.job.showUrl(), this.job.name())
+            };
+        });
+
+        context("without workspace", function() {
+            it("has the right header html", function() {
+                this.presenter.options.displayStyle = ["without_workspace"];
+                expect(this.presenter.headerHtml().toString()).toMatchTranslation(
+                    "activity.header.JobCreated.without_workspace", this.translation_params
```
kevin-alpine: I think this should be `activity.header.JobDeleted.default` ...

kevin-alpine:
* I confirmed that the rspec tests are all passing with this command: `rspec spec/models/job_spec.rb spec/models/events/job_deleted_spec.rb spec/models/events/job_created_spec.rb `

* I confirmed that, if I create / delete a job, I see it in the Workspace's activity stream, and I can comment on this:
alpine chorus-5

public/messages/Messages_en.properties
```diff
@@ -278,6 +278,13 @@ activity.header.WorkspaceChangeName.default={{actorLink}} updated the name of wo
 activity.header.TableauWorkbookPublished.default={{actorLink}} published {{datasetType}} {{datasetLink}} to the Tableau workbook {{tableauWorkbookLink}} in project {{tableauProjectLink}}
 activity.header.TableauWorkfileCreated.default={{actorLink}} added file {{workfileLink}} based on the {{datasetType}} {{datasetLink}} to workspace {{workspaceLink}}
 
+activity.header.JobCreated.default={{actorLink}} created job {{jobLink}} in workspace {{workspaceLink}}
```
kevin-alpine: Ok -- I think I found a problem here: I tried to get this string to appear.  If I look at the activity from inside the workspace, (http://localhost:3000/#/workspaces/4) I expect to see the `without_workspace` version of this string.

But, if I look at the activity from the root of the app, I expect to see the `default` version to appear and I expect to see a link to the workspace.  But, I do not.

alpine chorus-6

I worked for 1.5 hours to try and figure out why this is not happening!  :)  

(Just to put this out there, I may have broken something with your code when I teased it apart into 4 PRs.  I tried to check for this, but, I might be wrong.)

@atul-alpine can you confirm you see the same behavior as I do and you agree it's wrong?  

If you look at this line in activity_presenters.js: `} else if (self.workspace().id && self.get("actionType") !== "NoteOnWorkspace") {`

... on that line, for a WorkfileCreated event, `self.workspace().id` returns an integer.  But, for a JobCreated event, it returns `undefined` ... this is as far as I have been able to concretely get in understanding why it's going wrong.  There's something wrong with the ORM association (in JS) between the Job and the Workspace, I believe.

spec/javascripts/presenters/activity_presenters_spec.js
atul-alpine: Yeah right fixing it.

public/messages/Messages_en.properties
atul-alpine: It was working fine for me. Earlier code in Events::JobDeleted  was  has_targets job,:workspaces as I checked the targets column was not getting updated in DB and I was seeing the result.  During refactoring I removed workspace as target and kept it as activity since I was able to see workspace_id in events table.

So during refactoring I removed  has_targets :workspaces. I was still able to see the desired result.   When I explored more into has_targets I saw this 

def self.has_targets(*target_names)
      options = target_names.extract_options!
      self.target_names = target_names
      self.attr_accessible(*target_names, :as => [:create])

      target_names.each_with_index do |name, i|
        alias_getter_and_setter("target#{i+1}", name, options)
      end

      alias_method("primary_target", target_names.first)
    end

    def self.alias_getter_and_setter(existing_name, new_name, options)
      # The events table has a dedicated 'workspace_id' column,
      # so we don't alias :workspace to :target1 or :target2.
      # Subclasses should still specify the workspace as
      # a target if they need the workspace to be included
      # in their JSON representation.
      return if new_name == :workspace

      alias_method("#{new_name}=", "#{existing_name}=")
      alias_method(new_name, existing_name)
    end

Do read the comment specifically about workspace_id and how they don't set :target1. So what was happening was because I already had workspaces as targets in my backend model it was showing up until I created another milestone which in turn would change the JSON and will NOT set the getter for Event::JobDeleted. 

So @kevin-alpine you were right in detecting the ORM association was wrong but it was not on JS side but it was wrong on backend side. Funny thing is I could see "workspace" in the returned JSON object for all the activities (check attachment below) and was really struggling to find out what is wrong at the JS side; until I started thinking what changes I made during my refactoring.  I think backend and front end are really tightly bound and some special cases like these are really confusing. I will be committing the changes soon. 

screen shot 2015-07-15 at 11 20 47 am

app/models/job.rb
atul-alpine: Since we are calling those methods with current_user
```ruby
def create_job_created_event
    Events::JobCreated.by(current_user).add(
        :job => self,
        :workspace => workspace,
    )
  end

  def create_job_deleted_event
    Events::JobDeleted.by(current_user).add(
        :job => self,
        :workspace => workspace,
    )
  end
```

atul-alpine:
@kevin-alpine @prakash-alpine A note on separate PRs.

 By the way as a side note as a result of dividing these commits I faced some problems in testing the changes. For example my activity table had activities for milestones and when I started testing my changes I saw all the errors for milestone activity and error message was really really confusing. Then I realised milestone_created and milestone_updated models were removed. I realised this when I applied my stash for the fix I created when I fixed this issue on my original branch 11396. That stash had changes for both milestone and jobs model and they failed when I applied them. So had to test for one feature in one branch and another in different. Please note I had tested them in my original branch as well.
This would have happened even if I had started working on this particular branch. Issue is Database has no idea of what is going on (separate PRs etc) so it becomes inconsistent and we start to see all sorts of different errors. I think we need a better way to generate PRs. I think its better we keep features intact instead of dividing them based on Jira tickets. Does that makes sense or am I missing something?

kevin-alpine:
> As a result of dividing these commits I faced some problems in testing the changes.

I absolutely understand that there's additional overhead in dividing up your single branch, into 4.  But, as a result of divvying them up and reviewing them, we have discovered >=5 issues: 

* In #77 we discovered the leftover localization string `help_n_support`
* In #78 we discovered the inconsistencies between the new and existing notifications: one is bolded and one is not.  Also, Michael gave us good feedback on the naming of the keys.
* In this one #79 we've discovered two broken Jasmine tests and a bug in the way the new activity feed was being displayed on the app's dashboard.
* I have yet to review #80 but I imagine we'll find some things there.

And this is totally expected!  I have almost never submitted a PR that hasn't benefitted from peer review.  But some of these things would have come up again for sure either via another bug, or failing Jasmine tests in the CI build.  

The Chorus code is difficult to work with, and we are trying to turn the ship around.  That's why we have to be careful: each PR that we merge in has to either keep the code quality the same, or improve it.

(I do think that #79 and #80 could have been done in a single branch.  But we're both new to this codebase so I took the conservative routing of dividing them.)


> Issue is Database has no idea of what is going on (separate PRs etc) so it becomes inconsistent and we start to see all sorts of different errors.

I totally understand this is awkward, but we need to be able to deal with getting our development database into the state it needs to be to do precise, small, pieces of work.  

What I did was to `rake db:drop db:create db:migrate` and then reload the dev database Prakash gave me, via this command `./postgres/bin/psql -p8543 -h127.0.0.1 -d chorus_rails_development -f chorus_db.sql `.  Later when I understood more what was going wrong, I deleted the non-existent Single Table Inheritance Event types from the database using the Rubymine database GUI.

kevin-alpine:
I see that the two commits you made, fixed the issues I found.  So, this branch now "Looks Good to Me": LGTM.  Thank you @atul-alpine!

public/messages/Messages_en.properties
kevin-alpine: @atul-alpine,

Yes -- I pulled your updated code and it is working correctly for me now.  I dug into the `has_targets` and the association of Events with Activities and ... wow, it's quite a thing.  Anyways, it's very good for me to learn about this, because it's very relevant to the work I am doing to divide the codebase into components.

> The events table has a dedicated 'workspace_id' column, so we don't alias :workspace to :target1 or :target2. Subclasses should still specify the workspace as a target if they need the workspace to be included in their JSON representation.

This definitely explains the behavior we're seeing.  Thanks for pointing this out!

atul-alpine:
Hey Kevin!

Thanks for insightful comments. I am learning a lot from it. You are right
the cost of finding those bugs earlier in the cycle is way better and if
that costs us some more work at database side then its justified since
effort is only of 2 people and not lot of people once bug gets filed and it
goes through the cycle of tracking fixing, reviewing and re-testing,
updating Jira etc!

Awesome! thanks for the inputs. Will improve upon the suggestions!

-Atul


On Thu, Jul 16, 2015 at 2:40 AM, kevin-alpine <notifications@github.com>
wrote:

> As a result of dividing these commits I faced some problems in testing the
> changes.
>
> I absolutely understand that there's additional overhead in dividing up
> your single branch, into 4. But, as a result of divvying them up and
> reviewing them, we have discovered >=5 issues:
>
>    - In #77 <https://github.com/Chorus/chorus/pull/77> we discovered the
>    leftover localization string help_n_support
>    - In #78 <https://github.com/Chorus/chorus/pull/78> we discovered the
>    inconsistencies between the new and existing notifications: one is bolded
>    and one is not. Also, Michael gave us good feedback on the naming of the
>    keys.
>    - In this one #79 <https://github.com/Chorus/chorus/pull/79> we've
>    discovered two broken Jasmine tests and a bug in the way the new activity
>    feed was being displayed on the app's dashboard.
>    - I have yet to review #80 <https://github.com/Chorus/chorus/pull/80>
>    but I imagine we'll find some things there.
>
> And this is totally expected! I have almost never submitted a PR that
> hasn't benefitted from peer review. But some of these things would have
> come up again for sure either via another bug, or failing Jasmine tests in
> the CI build.
>
> The Chorus code is difficult to work with, and we are trying to turn the
> ship around. That's why we have to be careful: each PR that we merge in has
> to either keep the code quality the same, or improve it.
>
> (I do think that #79 <https://github.com/Chorus/chorus/pull/79> and #80
> <https://github.com/Chorus/chorus/pull/80> could have been done in a
> single branch. But we're both new to this codebase so I took the
> conservative routing of dividing them.)
>
> Issue is Database has no idea of what is going on (separate PRs etc) so it
> becomes inconsistent and we start to see all sorts of different errors.
>
> I totally understand this is awkward, but we need to be able to deal with
> getting our development database into the state it needs to be to do
> precise, small, pieces of work.
>
> What I did was to rake db:drop db:create db:migrate and then reload the
> dev database Prakash gave me, via this command ./postgres/bin/psql -p8543
> -h127.0.0.1 -d chorus_rails_development -f chorus_db.sql. Later when I
> understood more what was going wrong, I deleted the non-existent Single
> Table Inheritance Event types from the database using the Rubymine database
> GUI.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Chorus/chorus/pull/79#issuecomment-121748537>.
>



-- 
Regards,
Atul Veer.

kevin-alpine:
@prakash-alpine -- this PR is ready to be merged.  Do you approve merging it in?  

In my experience, normally this moment is arrived at via consensus and, it's normally obvious enough, and someone / anyone pushes the button.  If there's any doubt it's best to double check with everyone.

prakash-alpine:
LGTM

PR #81, opened 2015-07-14
Roles groups permissions

prakash-alpine:
Performance optimization for speeding up seed_permissions rake task.


AndrewLngdn:
local specs pass, I will continue to monitor the jenkins builds after this merge

PR #85, opened 2015-07-16
Roles groups permissions

AndrewLngdn:
Add caching to notifications and temporarily disable scope filtering


prakash-alpine:
This will bring the QA server (204,etc.) back to normal performance.  We will fix/test the performance problem is a isolated environment to unblock QA.

PR #86, opened 2015-07-17
fixes DEV-11731 and developer role update issue

AndrewLngdn:


app/models/user.rb
```diff
@@ -183,6 +183,17 @@ def admin=(value)
 
   scope :developer, where(:developer => true)
 
+  def developer=(value)
+    write_attribute(:developer, value)
```
prakash-alpine: @AndrewLngdn  write_attribute will not save the value to database. You have to add self.save! OR use update_attribute.  See http://www.davidverhasselt.com/5-ways-to-set-attributes-in-activerecord-in-rails-3/

PR #88, opened 2015-07-17
Roles & Permissions performance improvements - 1

prakash-alpine:
This is for code review purpose only. Do not merge with Master.

### DIFF
--- a/app/mixins/permissioner.rb
+++ b/app/mixins/permissioner.rb
@@ -4,7 +4,8 @@
 
 module Permissioner
   extend ActiveSupport::Concern
-  attr_accessor :chorus_scope_id
+  #attr_accessor :chorus_scope_id
+  #@chorus_scope_id = nil
 
   included do
     # after_create :initialize_default_roles, :if => Proc.new { |obj| obj.class.const_defined? 'OBJECT_LEVEL_ROLES' }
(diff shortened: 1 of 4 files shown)


app/mixins/permissioner.rb
```diff
@@ -4,13 +4,41 @@
...
+    begin
+      if self.class.name != 'Events::Base' && self.chorus_scope == nil
+        Chorus.log_error("No chorus scope found for #{self.class.name}")
+      elsif self.class.name != 'Events::Base'
+        @attributes['chorus_scope_id'] = self.chorus_scope.id
+      else
+        Chorus.log_error("---- chorus scope does not exists for #{self.class.name} ----")
+        @attributes['chorus_scope_id'] = ChorusScope.default_chorus_scope.id
+      end
+    rescue Exception => e
```
AndrewLngdn: What kind of exceptions were you running into here?

AndrewLngdn:
Looks good to me

app/mixins/permissioner.rb
prakash-alpine: The Events:Base class does not have a real object in DB and hence it does not have the 'id' attribute which throws an exception.  I think this is the only place where I had to make an exception.

prakash-alpine:
Do not merge this with master yet. Deploy directly from this branch onto 10.0.0.165

PR #90, opened 2015-07-20
11637: Create "download logs" button

kevin-alpine:
[backbone rails] Changes for DEV-11637: Creating a dialog called "Help and Support" which allows the users to download a zip file of the most recent logs.

https://alpine.atlassian.net/browse/DEV-11637


lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
+require 'fileutils'
+require 'zip/zip'
+
+module ChorusLogger
```
kevin-alpine: This is not a good name: this code doesn't have to do with logging, but rather with generating a Zipfile of the logs.

I suggest: ChorusLogArchiver ... or really, there's no need to have "Chorus" in there at all ... so maybe just `LogArchiver` ...

lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    zipfile_name = 'system/logs/' + zipfile_name
+    zipfile_name
+  end
+
+  def delete_logs
+    Dir.glob('*.log').each do  |log_file|
+      FileUtils.rm(log_file)
+    end
+  end
```
kevin-alpine: I think it should *definitely not* delete the logs.  There's no harm in allowing the user to zip and download the logs as many times as they want.

Second problem is that the UI does not give any indication that the log files will be deleted.  What if the user accidentally deletes / misplaces the Zip file?

I suggest deleting this code that has to do with deleting, it will cause problems and not really get us any benefit.

config/routes.rb
```diff
@@ -203,6 +203,8 @@
 
   post 'download_data', :controller => 'data_downloads'
 
+  post 'download_logs' , :controller => 'data_downloads'
```
kevin-alpine: This should not be a POST, it should be a GET ...

Create -> POST
Read -> GET
Update -> PUT / PATCH
Delete -> DELETE

We shouldn't delete the log files (see other comment), and we should allow them to be zipped and downloaded any number of times ... this is definitely a READ operation.

app/controllers/data_downloads_controller.rb
```diff
@@ -1,7 +1,20 @@
...
+  include ChorusLogger
 
   def download_data
     send_data params[:content], :type => params[:mime_type], :filename => filename_for_download(params[:filename])
   end
+
+  def download_logs
+    cookies['fileDownload'] = 'true'
+    log_file = generate_log_file
+    logger.info "file name :" + log_file.to_s
```
kevin-alpine: This looks like a bit left over from development.  I don't think it is meaningful to have in the codebase long term.

lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
...
+      log_files << log_file
+    end
+    Dir.chdir('..')
+    log_files
+  end
+
+  def compress_logs(log_files)
+    Dir.chdir('public/system/logs')
+
+    zipfile_name =  (Time.now.to_s + ".zip").gsub(/\s+/, "_").gsub(/:/,"_")
```
kevin-alpine: In order to figure out what `.gsub(/\s+/, "_").gsub(/:/,"_")` does, you practically have to execute the code: it's really difficult to read.

Instead, I suggest something like this:

```
# alpine_logs_20150720150924.zip
zipfile_name =  "alpine_logs_#{Time.now.to_formatted_s(:number)}.zip")
```

... put the formatted filename in the comment so people can see what the filename will look like when generated.

lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def generate_log_file
+      Dir.chdir(Rails.root)
+      log_file = compress_logs(generate_logs)
+      delete_logs
+      return log_file
+  end
+
+  def generate_logs
+    FileUtils::mkdir_p 'public/system/logs/'
```
kevin-alpine: Why not put these files into the tmp directory?  `tmp/log_archiver` ...

I could see putting it in the public directory if you were having the client download the file from the webserver, but you are using `send_file`, so its location doesn't matter at all.

app/controllers/data_downloads_controller.rb
```diff
@@ -1,7 +1,20 @@
 class DataDownloadsController < ApplicationController
   include FileDownloadHelper
+  include ChorusLogger
 
   def download_data
     send_data params[:content], :type => params[:mime_type], :filename => filename_for_download(params[:filename])
   end
+
+  def download_logs
+    cookies['fileDownload'] = 'true'
```
kevin-alpine: I don't see this `cookies['fileDownload']` being used again -- what is it for?

lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
...
+    Dir.chdir("log")
+    log_files = Array.new
+    no_of_lines = ChorusConfig.instance['logging.no_of_lines']
+
+    Dir.glob('*.log').each do  |log_file|
+      system("tail -#{no_of_lines} #{log_file} > ../public/system/logs/#{log_file}")
```
kevin-alpine: I had a problem here, on my Mac I had some files with spaces in the filename ... I suggest you wrap the filenames in doublequotes, like this:

`system("tail -n #{no_of_lines} \"#{log_file}\" > \"../public/system/logs/#{log_file}\"")`

lib/chorus_logger.rb
```diff
@@ -0,0 +1,46 @@
...
+      log_file = compress_logs(generate_logs)
+      delete_logs
+      return log_file
+  end
+
+  def generate_logs
+    FileUtils::mkdir_p 'public/system/logs/'
+    Dir.chdir("log")
+    log_files = Array.new
+    no_of_lines = ChorusConfig.instance['logging.no_of_lines']
```
kevin-alpine: I would give this key a better name, something like: `logging.download_last_number_of_lines` ... verbose is fine in a key like this, so people understand what it does.  `no_of_lines` is really vague.

app/controllers/data_downloads_controller.rb
```diff
@@ -1,7 +1,20 @@
 class DataDownloadsController < ApplicationController
```
kevin-alpine: I don't think we should put this in the `DataDownloadsController` -- both actions download files, but otherwise I don't think they have very much to do with one another.

It's better to have two small, independent controllers, than to munge together unrelated things.

Perhaps we could call this: `LogArchiverController`

lib/chorus_logger.rb
kevin-alpine: Hi @atul-alpine,

I apologize, I think what you were doing here is just to delete the truncated log files.  I don't think you meant to delete the actual existing log files, but just the truncated files which you just generated.  That makes total sense.

If you look in the `11637-create-download-logs-button-kt-edits` branch, you can see I rewrote these methods to use absolute paths.  I think it will read clearer that way, instead of using `Dir.chdir` ...

KT

kevin-alpine:
Hi @atul-alpine,

Please see the other "KT edits" PR that I created: #91 

Based on that branch, it's working for me.

If you agree with #91, go ahead and merge it in and then, I think we can merge & deliver this PR.

Best,
Kevin

lib/chorus_logger.rb
atul-alpine: @kevin-alpine  Hmm didn't knew about chorus_logger.rb. Well I was not looking at it just one functionality. I was more or less looking at it as a library of things we need to with logs, which might be there in future. Hence the name.

lib/chorus_logger.rb
atul-alpine: @kevin-alpine yeah that makes sense. I meant to delete the truncated log files. I liked you approach of having separate directories for archiving and staging. However, I am not sure if tmp is the right place for staging. I think right place should be public folder which is easier for user to locate in future.

config/routes.rb
atul-alpine: Yeah right this is READ operation. I have been playing with the code as it was not working. It was get earlier.

lib/chorus_logger.rb
atul-alpine: This is much more readable. I agree.

app/controllers/data_downloads_controller.rb
atul-alpine: Well, this if you look at the functionality we are in essence downloading data (in the form of logs) from server, so I thought instead of creating new controller it makes sense to use current structures. I had crated another controller for the same, when I realised we have data_download_controller I reverted my changes.

app/controllers/data_downloads_controller.rb
atul-alpine: Since downloads were not working for me. I wanted to check if download has occurred or not. It was for debugging, which could be later user for other clean up tasks. But anyways not needed with your changes.

app/controllers/data_downloads_controller.rb
atul-alpine: Yeah as I said earlier I did had a separate controller for this functionality but reverted my changes as I discovered data downloads controller. I think LogArchiverController is rather narrow name; may be we can have a generic name like LogManagerController to incorporate other log operations which might be needed in future? e.g rotate, download previous logs. thats what I had thought about earlier. What do you think?

atul-alpine:
Sounds great!

atul-alpine:
@kevin-alpine I have made changes to add Alpine logs and to match new layout. Please review backend code. I still need to clean up the UI.

PR #91, opened 2015-07-20
KT Tweaks and Changes for: "11637: Create 'download logs' button"

kevin-alpine:
Hi @atul-alpine,

Like I did before, here is a small PR with some changes related to readability.  I didn't really change much with your code, just renamed and organized things a bit.

When I run this code, it works, in my dev build:

Does this work for you too?

Best,
Kevin


lib/log_archiver.rb
```diff
@@ -0,0 +1,52 @@
+require 'fileutils'
```
kevin-alpine: Another comment is, I'm not sure if we should put this in the `lib` directory.  Normally I agree this would be the place for it, but in this app these sorts of files are either in the Mixins or Services directory.  We should maybe explore putting it into the services directory.

lib/log_archiver.rb
```diff
@@ -0,0 +1,52 @@
...
+    zipfile_path
+  end
+
+  def truncate_logs_into_staging_directory
+    Dir.chdir("#{Rails.root}/log")
+
+    num_lines = ChorusConfig.instance['logging.archiver_truncate_number_of_lines']
+
+    log_files = Array.new
+    Dir.glob('*.log').each do |log_file|
```
kevin-alpine: If we can configure an array of all the logfile locations, including Alpine logs, we could consume that array here and make this more sophisticated.  I don't think it would be too hard.

atul-alpine:
@kevin-alpine I have moved log_archiver to services folder, its working fine. Please review.
 I think thats the right place for it. I remember services folder when I had gone through the code base earlier but had forgotten about it. Still getting myself familiar with the code-base :).

kevin-alpine:
That all sounds good, @atul-alpine -- I'm going ahead and merging this back into your main feature branch.  We can continue with dev in there now that we're on the same page.  Thanks for your review, and it's good to hear you confirm that the feature is working for you, too.

PR #92, opened 2015-07-22
Added uniqueness constraints to prevent duplicate entries in roles & …

prakash-alpine:
Changes to avoid duplicates entries in roles & permissions table.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -2,9 +2,11 @@ class Role < ActiveRecord::Base
 
   attr_accessible :name, :description
 
+  validates :name, :presence => true, :uniqueness => true
+
   has_and_belongs_to_many :users, :uniq => true
-  has_and_belongs_to_many :groups
-  has_many :permissions
+  has_and_belongs_to_many :groups, :uniq => true
+  has_many :permissions, :dependent => :destroy
   has_many :chorus_object_roles
   has_many :chorus_objects, :through => :chorus_object_roles
 
--- a/db/clear_permissions.rb
+++ b/db/clear_permissions.rb
@@ -1,13 +1,32 @@
 # Clear all tables related to roles, groups and permissions.
 puts 'Clearing data entries for roles, groups and permissions'
-Role.destroy_all
-Group.destroy_all
-Permission.destroy_all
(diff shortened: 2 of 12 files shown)


db/clear_permissions.rb
```diff
@@ -1,13 +1,13 @@
 # Clear all tables related to roles, groups and permissions.
 puts 'Clearing data entries for roles, groups and permissions'
-Role.delete_all
-Group.delete_all
-Permission.delete_all
-ChorusScope.delete_all
-Operation.delete_all
-ChorusClass.delete_all
+Role.destroy_all
```
AndrewLngdn: For some reason, this occasionally leaves the collaborator role. It should be taken care of by the pure SQL script in the other branch, so I will make sure the two branches merge correctly

prakash-alpine: Yes.I noticed that. I think it triggers a callback somewhere that add the default_role back as collaborator. I could not find the code where it does this.

app/models/role.rb
```diff
@@ -2,9 +2,11 @@ class Role < ActiveRecord::Base
 
   attr_accessible :name, :description
 
+  validates :name, :presence => true, :uniqueness => true
```
AndrewLngdn: With the current schema setup, this line won't work for workspace-level roles. Each workspace has the 5 or so workspace role entries in the roles table, which means there will be one 'WorkspaceManager' role in the table for each workspace, which will cause errors because of this validation

prakash-alpine: Let us review this to-gather.  I am not sure how we want to handle workspace level roles in the Role table. My be we need to add instance_id column to role table?

AndrewLngdn:
I managed to trigger the Admin thing. I think it happens because of the uncheck_admin_role function on the user model. When you clear the permissions in this branch, it removes all the roles which triggers uncheck_admin_role. Since self.admin == true but self.roles.include?(admin_role) is false because we deleted it, it will set admin to false and then save the user model. This will remove all admin information from the site, so when you go back and try to re-assign admin roles in seed_permissions, it won't make anyone admin.

PR #97, opened 2015-07-23
update development rake to give more information

AndrewLngdn:


kevin-alpine:
Hi Andrew,

Hi, hey: I'm trying to review some of these PRs so we can improve our "flow."

K

lib/tasks/development.rake
```diff
@@ -31,9 +31,11 @@ namespace :development do
   task :init_database => [:generate_database_yml] do
     root = Pathname.new(__FILE__).dirname.join("../..")
     postgres_port = `ruby #{File.join(root, 'packaging', 'get_postgres_port.rb')}`.chomp
-    next if root.join("postgres-db").exist?
 
-    `DYLD_LIBRARY_PATH=#{root}/postgres/lib LD_LIBRARY_PATH=#{root}/postgres/lib #{root}/postgres/bin/initdb -D #{root}/postgres-db -E utf8`
+    unless root.join("postgres-db").exist?
```
kevin-alpine: I see that you changed `next` to `unless` -- so, what's really happening here is that the `chorus_control.sh start postgres` will still be executed and, fail?

Is that how it "gives more information" -- can you give a little more explanation of how you arrived at this point ... what frustrations is it solving, etc?

Thanks,
KT

AndrewLngdn: This task basically initializes a postgres db in the folder 'postgres-db' and runs a bunch of other rake tasks. The old task would look for the folder 'postgres-db' and if it was there, it would silently quit the script. Now (as you mentioned) it will startup postgres and attempt to run the other scripts, which are idempotent and will print some error messages if they've already been run. It's not a perfect solution to some of the problems with this rake task but it's a little bit better than it was before

alpinebuilder:

PR #98, opened 2015-07-24
Cleanup -- delete rando file that got committed: "pteli_chorus.sql"

kevin-alpine:


kevin-alpine:
Hi @prakash-alpine -- hey can you confirm this file shouldn't be in the repo?

kevin-alpine:
I confirmed with Prakash that this is correct.  Merging it in.

PR #99, opened 2015-07-24
Rails 4.1.10 upgrade 

kevin-alpine:
The initial state of this PR contains all the changes necessary so that the specs will pass, for Chorus 5.5.   I'm working on the changes necessary for 5.6 and, the plan is to merge this in very soon after the Release_5.6 tag is created.


app/helpers/authority.rb
```diff
@@ -179,14 +179,14 @@ def self.account_owner(data_source, current_user)
...
-  
+
   def self.retrieve_roles(user)
     roles = user.roles.clone
+
     user.groups.each do |group|
       roles << group.roles if group.roles.empty? == false
     end
-    # PT: Andrew, why do we need this?
-    roles.reject!{|r| r.class.name == 'Array'}
```
kevin-alpine: This code throws an exception in Rails 4, because:

"Note that where(...) returns a relation, not an array like the old finders. If you require an Array, use where(...).to_a."

(There are a good number of `Enumerable` methods that throw exceptions when you use them on ActiveRecord collections in Rails 4.)

@AndrewLngdn -- I don't understand why we need it either -- can you help me understand if it is still necessary?  We can adapt it, or remove it.

kevin-alpine:
@jleealpine: in the rails-4.1.10-upgrade branch, the `worklet_result_shared.rb` file was throwing an exception, because, per Rails 4 stricter "Autoloading" rules, it needs to be nested in the `events` directory.  

... because of:

```
module Events
  class WorkletResultShared < Base
```

... so, I moved it in there.

This is not really a Rails 4 related fix, it should perhaps be backported to the 5.6 release.

app/helpers/authority.rb
kevin-alpine: @AndrewLngdn: can you check this out?  I am trying to get some attention on this PR before I merge it in.

alpinebuilder:

PR #100, opened 2015-07-24
(MERGE 1st) [Rails 4.1.10 upgrade] Collapse migrations into 2 files.

kevin-alpine:


kevin-alpine:
I don't think this needs to be tied to #99 -- closing and opening another to merge into master.

PR #101, opened 2015-07-24
Collapse migrations

kevin-alpine:
As part of the Rails 4 upgrade, Prakash and I are collapsing the hundreds of migrations, down into just a few.  This is a good thing to periodically do: it's an easy way to delete a lot of code from the project.

This will need to be redone, before merging in.  Migrations are being added all the time so, right after the 5.6 release goes out, I will move the last few migrations into the BaseSchema and double check to make sure that all is right.

This PR is just to show our intent to make this change within the next few weeks, and as a place for debate.


alpinebuilder:

PR #102, opened 2015-07-27
DEV-8655 Keep the activity list when quickstart is visible in workspace view

jmlewis:
Keep activity stream in workspace quick start view by adding it as a subview.


jmlewis:
@prakash-alpine Not sure if assigning is the way that you track who's doing the code review. I figure you're the right person for my first few diffs :)

Also, rake jasmine:fixtures is failing for me — I may still need to update/add tests as part of this diff.

jmlewis:
Have the specs running. Added a basic spec that checks for the presence of the subview container. Given that the subview itself is independently tested in its own spec is this sufficient?

If so I believe this is ready for review. I made the layout changes we discussed.

alpinebuilder:

mmilano:
from discussions, we decided to hold on this for now and revise the sequencing of the page elements

PR #103, opened 2015-07-28
DEV-11860 Render the header just once per session

jmlewis:
This solution is slightly clunky — it needs to work around the central assumption in the router that more or less the entire page gets completely rerendered on nav. Generally this isn't a great approach since it can be expensive and lead to visual artifacts like the one that motivated DEV-11860. One of the goals of the updated UI framework should be to better support partial rerenders.

Most of the changes in _header.scss are just moving up a tab level.


app/assets/javascripts/models/session.js
```diff
@@ -16,7 +16,7 @@ chorus.models.Session = chorus.models.Base.extend({
     },
 
     loggedIn: function() {
-        return this._user && this._user.get("id");
+        return this.user() && this._user.get("id");
```
jmlewis: Allows _user to get lazily instantiated if it hasn't done so before a loggedIn check.

app/assets/javascripts/router.js
```diff
@@ -115,14 +115,30 @@
...
             return function () {
                 var args = _.map(_.toArray(arguments), function(arg) {
                     return decodeURIComponent(arg);
                 });
+
                 var navFunction = function() {
                     chorus.PageEvents.off();
+
+                    self.renderHeader();
+                    if (pattern.indexOf(":workspaceId") !== -1) this.header.workspaceId = args[0];
```
jmlewis: This is meant to replicate the functionality in pages.js lines 111-112. Should we clear workspaceId if we nav to something that isn't part of the workspaces route?

alpinebuilder:

jmlewis:
Was working with Prakash on this. Going to abandon.

PR #105, opened 2015-07-28
Fix for [DEV-11876] Added check for nil owner ID to account for incon…

prakash-alpine:
Fix for [DEV-11876] Added check for nil owner ID to account for inconsistent database entries.  Existing customer database does not have strong validation and constraints.


AndrewLngdn:
looks fine. merging

PR #106, opened 2015-07-28
Fix for [DEV-11671] Attempting to attach a dataset to an Insight caus…

prakash-alpine:
Filtered out Proc object from cache.  Proc object is being used for uploaded files and does not need to be cached.

### DIFF
--- a/app/presenters/event_presenter.rb
+++ b/app/presenters/event_presenter.rb
@@ -109,7 +109,7 @@ class EventPresenter < Presenter
       datasets = model.datasets
       datasets.each do |dataset|
         model_hash = present(dataset, {:workspace => model.workspace}.merge(extended_options))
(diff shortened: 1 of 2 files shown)


midnighteuler:
If we merge the above it could lead to subsequent bugs that manifest due to omissions in the cached data.

I encountered this issue with caching workfiles as attachments in activities.
See this commit: https://github.com/Chorus/chorus/commit/f87147df3c6e44cc6b98cbb70d1f2112e6259897

The attribute that appeared to be a "Proc" was a method yielding serialized data. Presenting/caching using "attributes" on the object as in the above commit will (should) yield the object's associated data in a serializable way.

prakash-alpine:
@midnighteuler I am merging this PR so it can get into the next build tonight.  Please review anyways

PR #108, opened 2015-07-29
Fix for [DEV-11901] Site Snapshot widget not being updated.  Needed t…

prakash-alpine:
…o add default_group for newly created user. Also added after_destroy calls to delete HABTM associations table entries.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -25,12 +25,6 @@ class UsersController < ApplicationController
     developer_role = Role.find_by_name("Developer")
     User.transaction do
       user.save!
-      if user_params[:admin]
-        user.roles << admin_role unless user.roles.include? admin_role
-      end
-      if user_params[:developer]
-        user.roles <<  developer_role unless user.roles.include? developer_role
-      end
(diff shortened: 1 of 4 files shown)


app/controllers/users_controller.rb
```diff
@@ -21,11 +21,19 @@ def create
...
     user.developer = user_params[:developer] if user_params.key?(:developer)
+    admin_role = Role.find_by_name("Admin")
+    developer_role = Role.find_by_name("Developer")
     User.transaction do
       user.save!
-
-      Role.find_by_name("Admin").users << user if user_params[:admin]
-      Role.find_by_name("Developer").users << user if user_params[:developer]
+      if user_params[:admin]
+        user.roles << admin_role unless user.roles.include? admin_role
```
AndrewLngdn: Prakash, I've removed these admin_role and developer_role lines in master, I don't think they're necessary. Lines 21 and 22 will add the correct roles for admin and developer (we're using 'WorkflowDeveloper' now)

AndrewLngdn:
Other than that one comment the PR looks good. The lines I removed from master are probably what's causing the merge conflict

PR #111, opened 2015-08-01
Further work for DEV-11378 Hive Data Source entry screen

kevin-alpine:
This is to fix: https://alpine.atlassian.net/browse/DEV-11915

* Mike Souza helped me to figure out that, we need to also modify the 'connection_parameters' in the JSON API, as our application communicates with Hadoop in 2 ways: from Chorus -> Hadoop, and
from Chorus -> Alpine -> Hadoop.  Alpine gets the connection_parameters from the Chorus API.  So, I take care of this.

* We must also modify the UI to work around the fact that these fields are moved from connection_parameters and into their own columns.  I add "validations" to the form to make it so that these keys cannot be added as "connection_parameters" (because, otherwise we have a "source of truth" problem.)  I also add a migration to move any instances of these fields out of the connection_parameters column, and into their distinct columns.

### DIFF
--- a/app/assets/javascripts/dialogs/data_sources/hdfs_connection_parameters_dialog.js
+++ b/app/assets/javascripts/dialogs/data_sources/hdfs_connection_parameters_dialog.js
@@ -14,7 +14,11 @@ chorus.dialogs.HdfsConnectionParameters = chorus.dialogs.Base.extend({
     },
 
     setup: function () {
-        this.pairs = this.model.get('connectionParameters') || [{key: '', value: ''}];
+        this.pairs = this.model.connectionParametersWithoutHadoopHive();
+        if(this.pairs.length == 0) {
+          this.pairs = [{key: '', value: ''}];
+        }
+
         this.host_info = { host: '', port: 8088 };
     },
 
@@ -22,6 +26,7 @@ chorus.dialogs.HdfsConnectionParameters = chorus.dialogs.Base.extend({
         e && e.preventDefault();
 
         this.preservePairs();
+
         this.model.set('connectionParameters', this.pairs);
 
         if (this.validatePairs() === true) {
@@ -35,9 +40,15 @@ chorus.dialogs.HdfsConnectionParameters = chorus.dialogs.Base.extend({
         // Perform manual validation
         var validation_errors = {};
         for (var k in this.pairs) {
+            var key = this.pairs[k].key.trim();
+
             // Don't allow any key to be blank.
-            if (this.pairs[k].key.trim() === '') {
+            if (key === '') {
                 validation_errors['key_' + k] = t('validation.required', {fieldName: "Key"});
+            }
+
+            if (key === 'is_hive' || key === 'hive.metastore.uris') {
+              validation_errors['key_' + k] = t('data_sources.dialog.validation.hadoop_hive_keys_not_allowed');
             }
         }
 
--- a/app/assets/javascripts/models/abstract_data_source.js
+++ b/app/assets/javascripts/models/abstract_data_source.js
@@ -88,5 +88,23 @@ chorus.models.AbstractDataSource = chorus.models.Base.extend({
 
     usage: function() {
         return false;
+    },
+
+    // KT TODO refactor: the two methods following, belong more in HdfsDataSource -- but, it's difficult to get that
+    // to work, due to the way that the DataSourceNewDialog is written.
+    numberOfConnectionParameters: function() {
+        return this.connectionParametersWithoutHadoopHive().length;
+    },
+    connectionParametersWithoutHadoopHive: function() {
+      var pairs = this.get('connectionParameters');
+
+      pairs = _.map(pairs, function (pair) {
+        if(pair['key'] != 'is_hive' && pair['key'] != 'hive.metastore.uris') {
+          return pair;
+        }
+      });
+      pairs = _.compact(pairs);
+
+      return pairs;
     }
 });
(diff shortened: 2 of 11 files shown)


kevin-alpine:
@midnighteuler please review!  It's mostly all that we discussed.  The only problem is, I realized ... unless we forcibly exclude the "hive hadoop" keys from the "connection parameters" dialog, they will leak in their right away.  If we allow that to happen, there will be two places to configure the keys, and only the inner one (in the connection parameters screen) will take.  This can't work.

app/assets/javascripts/models/hdfs_data_source.js
```diff
@@ -5,6 +5,11 @@ chorus.models.HdfsDataSource = chorus.models.AbstractDataSource.extend({
     shared: true,
     entityType: "hdfs_data_source",
 
+    beforeSave: function() {
+      this._super("beforeSave");
+      this.set('connectionParameters', this.connectionParametersWithoutHadoopHive());
+    },
```
kevin-alpine: @midnighteuler: I think this is the most questionable part of the PR -- what it is, is just an attempt to fix a bug.  For some reason, if you edit a Hive Hadoop Data Source, when you save it, you see the 'is_hive' and 'hive.metadata.uris' key duplicated 5 times!

These are the same keys which I am trying to delete.  I imagine I'll figure it out on my own eventually but ... I have to leave work, and I'm a bit mystified at this point.  Classic stuff.

So, this is one big pending issue and we can't merge in this PR till I figure this out.  Do you have any insight?

alpine chorus

kevin-alpine: Another thing we could do is, to create a setter in the Rails model which will filter these keys out.  But, to do that seems like, we're getting sloppy and it feels stinky.

kevin-alpine: While biking I realized this must be coming from the new method in hdfs_data_source -- where I do:

collection_parameters << hive_metastore_location

I should use a local var there, and not the attr itself. The model might be getting saved with the extra params merged in.

kevin-alpine: I pushed up a fix for this, it was what I had guessed.

app/assets/javascripts/models/abstract_data_source.js
```diff
@@ -88,10 +88,5 @@ chorus.models.AbstractDataSource = chorus.models.Base.extend({
 
     usage: function() {
         return false;
-    },
-
-    numberOfConnectionParameters: function () {
-        var connectionParams = this.get('connectionParameters');
-        return connectionParams ? connectionParams.length : 0;
```
kevin-alpine: @midnighteuler -- I didn't think this belonged in `abstract_data_source` because, the "connection parameters" only have to do with the HdfsDataSources.  So, I moved it into `hdfs_data_source.js` and modified it suit my needs.

app/assets/javascripts/models/hdfs_data_source.js
```diff
@@ -59,6 +59,19 @@ chorus.models.HdfsDataSource = chorus.models.AbstractDataSource.extend({
...
+
+    connectionParametersWithoutHadoopHive: function() {
+      var pairs = this.get('connectionParameters');
+
+      pairs = _.map(pairs, function (pair) {
+        if(pair['key'] != 'is_hive' && pair['key'] != 'hive.metastore.uris') {
+          return pair;
+        }
+      });
+      pairs = _.compact(pairs);
```
kevin-alpine: This part is perhaps not very obvious ... if pairs is null, it remains null until this line, where `_.compact(pairs);` returns `[]` ...

app/assets/javascripts/dialogs/data_sources/hdfs_connection_parameters_dialog.js
```diff
@@ -14,7 +14,11 @@ chorus.dialogs.HdfsConnectionParameters = chorus.dialogs.Base.extend({
     },
 
     setup: function () {
-        this.pairs = this.model.get('connectionParameters') || [{key: '', value: ''}];
+        this.pairs = this.model.connectionParametersWithoutHadoopHive();
+        if(this.pairs.length == 0) {
+          this.pairs = [{key: '', value: ''}];
+        }
```
kevin-alpine: @midnighteuler asks: "are those comparisons equivalent?" ... meaning `this.model.get('connectionParameters') || [{key: '', value: ''}]` vs `if(this.pairs.length == 0)`

Good question -- I found the old version to not work in the case of an empty array ... I don't know, it may never have been empty in reality, but `[] || [{key: '', value: ''}]` returns `[]` ... which isn't the intent.  That's why I think the `if(this.pairs.length == 0)` is better.

kevin-alpine:
Hi @midnighteuler: I also added a test case to `/models/hdfs_data_source_spec.rb` for the connection parameter change in the case of hive, as you requested.  For the Backbone models issue that we discussed, I ended up moving all the new methods, back to `AbstractDataSource`.  I was able to change the backing model of `DataSourceNewDialog` to `AbstractDataSource` ... so I think that helps a little.

Please give me your LGTM when you get a chance, so that I can merge this in.

Thanks!
KT

app/assets/javascripts/models/abstract_data_source.js
kevin-alpine: Ultimately I had to replace this stuff in `abstract_data_source` ...

PR #112, opened 2015-08-04
[Dev-12001] - SiteSnapshot numbers are doubled if user is created after an upgrade but before migrate_permissions

prakash-alpine:
Needed add unique constraint for user and group HABTM relationship to ensure only unique entries are returned.


AndrewLngdn:
LGTM

PR #114, opened 2015-08-04
DEV-12000 update authority to bypass permissions on hdfs data source

AndrewLngdn:


spec/helpers/authority_spec.rb
```diff
@@ -1,57 +0,0 @@
-require File.dirname(__FILE__) + '/../spec_helper'
```
AndrewLngdn: these tests were out-dated and the specs weren't actually running them

prakash-alpine:
LGTM

PR #115, opened 2015-08-04
fix User#admin= so it can accept strings in order to unbreak Robot testing

AndrewLngdn:


app/models/user.rb
```diff
@@ -41,58 +41,9 @@ class User < ActiveRecord::Base
 
   # roles, groups, and permissions
   has_and_belongs_to_many :groups, :uniq => true
-  has_and_belongs_to_many :roles, after_add: :check_admin_role, after_remove: :uncheck_admin_role, :uniq => true
+  has_and_belongs_to_many :roles, :uniq => true
```
prakash-alpine: @AndrewLngdn 
Why did you remove after_add and after_remove callbacks?

AndrewLngdn: They were unnecessary and were causing strange behavior when the automation was trying to create and modify users. I don't think we will need them since role addition is taken care of by admin=, but if we decide that we need them, I can write specs around them and re-implement the functionality

app/models/user.rb
```diff
@@ -163,36 +124,38 @@ def self.named(username)
...
+    admin_role = Role.find_by_name("Admin")
+    app_manager_role = Role.find_by_name("ApplicationManager")
+    site_admin_role = Role.find_by_name("SiteAdministrator")
+
+    if value == true || value == "true"
+
+      admin_role.users << self unless admin_role.users.include? self
+      app_manager_role.users << self unless app_manager_role.users.include? self
+      site_admin_role.users << self unless site_admin_role.users.include? self || self.username != chorusadmin
       write_attribute(:admin, value)
```
prakash-alpine: Will write_attribute(:admin, "true") will work? I am not sure..

AndrewLngdn: Good question, it does work, Rails/the DB converts it to a boolean. That was actually part of the reason the original bug existed. 

Just to be sure, I've added a spec for it to user_spec.rb

prakash-alpine:
LGTM

PR #116, opened 2015-08-05
DEV-11947: Worklets catalog page

kevin-alpine:
Hello team!

I worked on https://alpine.atlassian.net/browse/DEV-11947 for most of today and made a lot of progress.

I'm putting my progress in a PR so that I can get @jleealpine's & @midnighteuler's feedback.  Do you agree with the way I'm doing things so far?

@mmilano: some of the screens do not look exactly like your designs.  This is because, your comps don't match the way that the app currently looks.  I'm using the abstracted "List Heading Bar" (which is already used on a number of pages) and I don't think I should / can restyle it to match your comps, at this point.  Please let me know if you disagree.

There's nothing in the app currently like "tiles" so, tomorrow I will continue fine-grained styling of the tiles and, should be able to make them closely match your designs.

Here's a screenshot of what it currently looks like:
worklets_catalog_page_first_day

Looking forward to continuing tomorrow -- best,
KT


kevin-alpine:
@mmilano: I just pulled a "hexagonal image" off the internet.  Do you have a more correct default image(s) that I can use?

mmilano:
@kevin-alpine - mike already implemented the default avatar image for worklets; check with him on using that, and how to access/reference it.

mmilano:
@kevin-alpine: yes, top elements should match site per the standard behaviors. 
however:
* put # of worklets on the left
* change the placeholder text for the filter to simplified version

the existing 'tile' like behavior occurs on the home page dashboard widget with the project cards.

mmilano:
@kevin-alpine fyi, the existing worklet avatars can be seen when editing a worklet in a workspace, on the "configure details." at the moment, the page first loads blank - you have to re-select the page from the drop down. looks like attached.
screen shot 2015-08-04 at 10 29 08 pm

app/assets/javascripts/views/workfiles/published_worklet_view.js
```diff
@@ -0,0 +1,13 @@
+chorus.views.PublishedWorklet = chorus.views.Base.include(chorus.Mixins.TagsContext).extend({
```
midnighteuler: I think using the "item" naming, or maybe "card", or (...) would be more clear here as "PublishedWorklet" is in use elsewhere.

app/assets/javascripts/views/workfiles/published_worklet_list_view.js
```diff
@@ -0,0 +1,28 @@
...
+                    model: model
+                });
+                this.published_worklets.push(view);
+                this.registerSubView(view);
+                div.append(view.render().el);
+            } catch (err) {
+                chorus.log(err.message, err, "processing published_worklet", model);
+                if (chorus.isDevMode()) {
+                    var action, id;
+                    try {action = model.get("action");  id = model.id;} catch(err2) {}
```
midnighteuler: The exception handling strategy seems a bit cluttered and possibly unnecessary.

app/assets/javascripts/views/workfiles/published_worklet_list_view.js
```diff
@@ -0,0 +1,28 @@
...
+        this.published_worklets = [];
+
+        this.collection.each(function(model) {
+            try {
+                var view = new chorus.views.PublishedWorklet({
+                    model: model
+                });
+                this.published_worklets.push(view);
+                this.registerSubView(view);
+                div.append(view.render().el);
```
midnighteuler: You might see better performance by creating each model's subview in a view preRender method, and then use document.createDocumentFragment() in postRender; something like:
```
            var container = document.createDocumentFragment();
            _.each(this.collection.models, function(model) {
                container.appendChild((model's preRenderedView).render().el);
            });
            this.$el.append(container);
```

You can see an example of this approach in: chorus/app/assets/javascripts/views/dashboard/dashboard_project_list_view.js

app/assets/stylesheets/components/_card.scss
```diff
@@ -0,0 +1,33 @@
...
+
+  min-height: 160px;
+  position: relative;
+  vertical-align: top;
+  display: inline-block;
+  margin: 0px 0px 10px 10px;
+  border: $border-canon;
+  @include border-radius($card-border-radius);
+  overflow: hidden;
+}
```
kevin-alpine: @mmilano: Per your comment, I found the `card` behavior on the `project_dashboard` and abstracted it into this `card` component, which is used on both the dashboard, and the Worklets Catalog page.

kevin-alpine:
@midnighteuler: I ended up deleting a lot of the custom views I had created and just using `MainContentList` more fully -- that is why your earlier comments are now outdated.

@mmilano: I am now using a "simplified placeholder text for the filter"

app/assets/javascripts/views/workfiles/published_worklet_list_view.js
```diff
@@ -0,0 +1,16 @@
...
+        return {
+            url: model.showUrl()
+        };
+    },
+
+    postRender: function() {
+        // KT: a hack to adapt list_content_details.hbs to our needs:
+        $(".list_content_details").find(".explore, .button_holder").remove();
+        $(".list_content_details").find(".count").detach().prependTo('.list_content_details');
+    }
```
kevin-alpine: @mmilano: Re: "put # of worklets on the left" -- I'm able to do this at the cost of this hack.  It's not as bad as it first appears though because the 'explore' div adds nonsensical padding ...

Without this hack, it looks like this:

need_minor_hack

So, I either need to add complexity to ListContentDetails, or do this hack.

kevin-alpine:
Ok guys, I am pretty comfortable with this PR now.  I would like to merge it in soon so that QA can take a look at this page earlier rather than later.  

@midnighteuler: I have the ticket for "Create ability to edit Worklet images" assigned to me in a separate ticket: https://alpine.atlassian.net/browse/DEV-11828 and I suggest that I do that in a separate PR so that we can merge this one in earlier.

Now it looks like this:

alpine chorus

app/assets/javascripts/views/workfiles/published_worklet_list_view.js
```diff
@@ -0,0 +1,16 @@
...
+
+    collectionModelContext: function (model) {
+        return {
+            url: model.showUrl()
+        };
+    },
+
+    postRender: function() {
+        // KT: a hack to adapt list_content_details.hbs to our needs:
+        $(".list_content_details").find(".explore, .button_holder").remove();
```
kevin-alpine: Chatted with Mike S and he suggests just doing this w/ CSS.

app/models/worklet.rb
```diff
@@ -6,6 +6,13 @@ class Worklet < AlpineWorkfile
 
   has_many :variables, :foreign_key => :workfile_id, :class_name => 'WorkletVariable'
 
+  attr_accessible :worklet_image
+  has_attached_file :worklet_image, :styles => { :icon => "50x50>" },
+                    :path => ":rails_root/system/:class/:id/:style/:basename.:extension",
+                    :url => "/:class/:id/worklet_image?style=:style",
+                    :default_url => "/images/worklet_images/default.png"
+  validates_attachment_content_type :worklet_image, :content_type => /\Aimage\/.*\Z/
+
```
kevin-alpine: Chatted w/Mike S and he suggests moving this Paperclip attachment business into its own PR for https://alpine.atlassian.net/browse/DEV-11828.

PR #117, opened 2015-08-05
update app controller to redirect to login with invalid authenticity token

AndrewLngdn:
@midnighteuler can you think of any adverse effects from this? It's reverting back to default rails behavior

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,7 +15,6 @@ class ApplicationController < ActionController::Base
   before_filter :set_collection_defaults, :only => :index
   before_filter :extend_expiration
   rescue_from 'ActionController::MissingFile', :with => :render_not_found
-  rescue_from 'ActionController::InvalidAuthenticityToken', :with => :invalid_authenticity_token
   rescue_from 'ActiveRecord::RecordNotFound', :with => :render_not_found
   rescue_from 'ActiveRecord::RecordInvalid', :with => :render_not_valid
   rescue_from 'ApiValidationError', :with => :render_not_valid
@@ -54,10 +53,6 @@ class ApplicationController < ActionController::Base
 
   def verified_request?
     super || params[:session_id]
-  end
-
-  def handle_unverified_request
-    raise ActionController::InvalidAuthenticityToken
   end
 
   def set_current_user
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -15,7 +15,6 @@ class ApplicationController < ActionController::Base
   before_filter :set_collection_defaults, :only => :index
   before_filter :extend_expiration
   rescue_from 'ActionController::MissingFile', :with => :render_not_found
-  rescue_from 'ActionController::InvalidAuthenticityToken', :with => :invalid_authenticity_token
```
prakash-alpine: @AndrewLngdn  I forgot to review this PR earlier. I am not clear on the impact of this change. Let us discuss live.

AndrewLngdn: All the jasmine and rails specs pass, but it's hard to say what will happen with the chorus/alpine integration. Since we don't have a way of locally testing the integration, we would have go through the entire jenkins build process and wait for the automation tests. 

On the purely Chorus side of things, I don't see a problem with this commit. The old behavior was very unintuitive: POST/PUT/DELETE commands would fail with no error messages to the frontend. 

The new behavior doesn't allow any new functionality, it just kicks the user back to the login page when they have an invalid token

prakash-alpine:
LGTM. Please merge with master and look out for any bugs coming from QA.

app/controllers/application_controller.rb
```diff
@@ -56,10 +55,6 @@ def verified_request?
     super || params[:session_id]
   end
 
-  def handle_unverified_request
-    raise ActionController::InvalidAuthenticityToken
-  end
-
```
prakash-alpine: @AndrewLngdn  Why did you remove handle_unverified_request method?

AndrewLngdn: @prakash-alpine handle_unverified_request is a Rails method that is called when Rails detects an invalid request token. It was being over-written, and instead of calling the Rails method that deletes the session and logs the user out (desired behavior), it was raising that ActionController error which would return a 403 without logging the user out

prakash-alpine: OK. Thanks for the explanation. LGTM.

prakash-alpine:
LGTM. Good to go.

PR #118, opened 2015-08-06
DEV-11828: Create ability to edit Worklet images

kevin-alpine:
You need to have an Alpine dev environment setup to access the pages where the avatar editing takes place, so I am collaborating with @jleealpine to get the Rails half taken care of, he
will handle the Backbone bits.

https://alpine.atlassian.net/browse/DEV-11828


app/assets/javascripts/templates/published_worklet_list.hbs
```diff
@@ -5,7 +5,7 @@
 {{#if models.length}}
 
   {{#each models}}
-    <div class="card" style="background-image: url('{{ avatarUrl }}');">
+    <div class="card" style="background: linear-gradient(rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0.2)), url('{{ avatarUrl }}');">
```
kevin-alpine: @jleealpine: since the design has white text over a user-uploadable image, there's a good chance the users will sometimes upload very light images ... so, this darkens the images slightly so that the text remains readable.

alpine chorus-5
alpine chorus-2

Reference:
http://stackoverflow.com/questions/23208200/how-to-darken-a-background-using-css

app/controllers/worklets_controller.rb
```diff
@@ -49,8 +49,27 @@ def update
...
+
+  def tmp_update_image
+    @workspace = Workspace.find(params[:workspace_id])
+    @worklet = Worklet.find(params[:id])
+
+    @worklet.attributes = params[:published_worklet]
+    @worklet.save
+
+    redirect_to tmp_edit_image_workspace_worklet_path(@workspace, @worklet)
+  end
```
kevin-alpine: I stubbed out a simple ERB based form to do this dev, so that I could upload images and test the #image action.  This may be useful to you.  You can access the form using URLs of this form: http://localhost:3000/workspaces/35/worklets/632/tmp_edit_image

We need to delete these methods before merging to master.

app/views/worklets/tmp_edit_image.erb
```diff
@@ -0,0 +1,4 @@
+<%= form_for @worklet, :url => tmp_update_image_workspace_worklet_path(@workspace, @worklet), :html => { :multipart => true } do |form| %>
+  <%= form.file_field :worklet_image %>
+  <%= form.submit %>
+<% end %>
```
kevin-alpine: We need to delete this file before merging to master.

config/routes.rb
```diff
@@ -109,6 +109,8 @@
     end
     resources :worklets, :only => [:create, :show, :update, :destroy] do
       member do
+        get 'tmp_edit_image'
+        put 'tmp_update_image'
```
kevin-alpine: We need to delete these temporary routes before merging to master.

app/controllers/worklets_controller.rb
```diff
@@ -49,8 +49,27 @@ def update
...
+
   def image
-    send_data GeoPattern.generate(worklet.id, generator: GeoPattern::HexagonPattern), type: 'image/svg+xml', disposition: 'inline'
+    if worklet.worklet_image.present?
+      send_file worklet.worklet_image.path(:display), :type => worklet.worklet_image_content_type, disposition: 'inline'
+    else
+      send_data GeoPattern.generate(worklet.id, generator: GeoPattern::HexagonPattern), type: 'image/svg+xml', disposition: 'inline'
+    end
```
kevin-alpine: @midnighteuler & I discussed this ... the worklet_images can always be accessed at the same URL of the form `/workspaces/35/worklets/632/image` ... and here we just ...

1) if a custom image has been uploaded, use `send_file`
2) if no custom image has been uploaded, GeoPattern always generates the same image for the same random seed, and we're using the model ID as the seed, so we can just generate the default image using GeoPattern

kevin-alpine:
#### Backbone beginnings

@jleealpine: you're more familiar with this than I am, but @midnighteuler and I thought it'd be best to model the Backbone half off of what already exists for uploading User profile images.  

I was looking at [this spot](https://github.com/Chorus/chorus/blob/master/app/assets/javascripts/views/users/user_edit_view.js#L21) as a good place to start.

app/assets/javascripts/templates/published_worklet_list.hbs
kevin-alpine: @mmilano: any comments about "Worklets Catalog design has white text over a user-uploadable image" and my proposed solution?

When looking at your designs it looked like you were already doing something like this.

mmilano: it is a fine general solution.
you really dont need the gradient, however (a gradient requires more render time and effort and memory) since a solid would work.

easier might be to simply use the css brightness filter, though.
examples: http://css3.bradshawenterprises.com/filters/

just please dont use inline css - create external css rules for this.

kevin-alpine:
Jason incorporated these changes into https://github.com/Chorus/chorus/commit/68a9754bca45d12df54d1dba7f481b4898c97de0 ... so, I am closing this PR as it's effectively already been merged.

PR #120, opened 2015-08-11
Fixes DEV-11959 and DEV-12039

AndrewLngdn:


prakash-alpine:
@AndrewLngdn Make sure you test fix for DEV-11959 thoroughly in your local environment. Particularly for non-chorusadmin users the numbers in dashboard (snapshot), workspaces, activities, should not be affected by this change

AndrewLngdn:
@prakash-alpine are there tests around the dashboard snapshot? It works locally, but it is unsustainable to be testing this by hand every time we make a change

prakash-alpine:
LGTM.

PR #121, opened 2015-08-12
Fixes DEV-11966 by adding explicit app level user role and removing object-level user role

AndrewLngdn:


prakash-alpine:
LGTM. Merging it.

PR #124, opened 2015-08-13
DEV-12109 update user.rb default roles to include User role

AndrewLngdn:
Also fixes a bug with the 'User' permissions not being defined in seed_permissions.rb


prakash-alpine:
LGTM.  Merging it myself as @AndrewLngdn  is on vacation.

PR #125, opened 2015-08-13
For Paperclip image processing, use Imagemagick, not imgscalr.jar

kevin-alpine:
Reverting Paperclip to use its standard image processing library: Imagemagick ...
* adding statically compiled versions of the utilities we need (convert & identify)
* adding the build script
* removing imgscalr.jar

DEV-11668: Explore including ImageMagick binaries in Chorus packaging & installer scripts
https://alpine.atlassian.net/browse/DEV-11668


packaging/build-imagemagick.sh
```diff
@@ -0,0 +1,55 @@
...
+    mkdir -p ../../../vendor/imagemagick/$os_friendly_name
+    cp utilities/convert ../../../vendor/imagemagick/$os_friendly_name
+    cp utilities/identify ../../../vendor/imagemagick/$os_friendly_name
+
+  popd
+popd
+
+rm -rf build
+
+echo "Imagemagick built for $os_friendly_name."
```
kevin-alpine: @holdenk: We chatted briefly about this several weeks ago.  This is one of the last big "Rails 4 upgrade" issues, so I've been working on it this week and I've come up with this PR.

I'm not a C programmer at all -- I'd basically like a "sanity check" of this code, assuming it works, is it reasonable?  If you have time, could you help me with this?

@mmilano: I saw you updated the `build-nginx.sh` script in May -- are you familiar with this sort of stuff?  I'd love a code review if you have time.

Thanks!
KT

mmilano:
@kevinmtrowbridge, afraid that everything i did in that file = trial + error + google.
the nginx was pretty simple; i only had to modify the existing.

mmilano:
@kevinmtrowbridge you might look at the postgres file too. notice how many builds of it there are.

kevin-alpine:
While testing this branch, I confirmed it works in OSX.  I also am pretty sure the Imagemagick binaries are working on CentOS 6.6.  However, when testing uploading a profile picture in the packaged / installed build, I get an exception.

This is the same exception that Prakash & Andrew discovered a few weeks ago when we had upgraded the `rack` gem version from 1.4.5 to 1.4.7: `undefined method `multipart_part_limit' for Rack::Utils:Module`

At that time Prakash and I had realized more about how `jetpack` works with `jruby-rack` to basically, allow Rails to run inside of Jetty.  Our stack works something like this, as far as I can tell: rails -> rack -> jruby-rack -> jetty

So, basically ... this branch is blocked by a totally different issue, that ... for Rails 4 we need `rack (~> 1.5.2` ... and seemingly if we upgrade it past 1.4.5 it breaks file uploads.

packaging/build-imagemagick.sh
holdenk: I'd add --enable-static=yes from looking at the documentation but if the current build works thats probably ok too. But from a quick sanity check point of view this looks reasonable.

alpinebuilder:

PR #129, opened 2015-08-14
Fixed DEV-11057 bug relating to deleting tags

curran:
With this fix, the behavior when deleting the first or last tag in the tag list is that the sidebar becomes blank, which is the correct behavior. When deleting a tag other than the first one, the behavior is that the next tag in the list becomes selected (same behavior as before).

Ideally, the sidebar should always become blank after deleting a tag, but I could not figure out how to get this behavior. This is related to the "selectedIndex" of the tag list not being updated after deleting a tag.

### DIFF
--- a/app/assets/javascripts/pages/tags/tag_index_page.js
+++ b/app/assets/javascripts/pages/tags/tag_index_page.js
@@ -14,6 +14,11 @@ chorus.pages.TagIndexPage = chorus.pages.Base.extend({
 
         this.handleFetchErrorsFor(tags);
         this.listenTo(tags, "loaded", this.render);
-        this.listenTo(tags, "remove", this.render);
+
+        this.listenTo(tags, "remove", function (tag) {
+          chorus.PageEvents.trigger('clear_selection', tag);
+          this.render();
+        });
+
     }
 });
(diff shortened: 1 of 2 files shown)


prakash-alpine:
@curran What do you mean by this comment
"Ideally, the sidebar should always become blank after deleting a tag, but I could not figure out how to get this behavior."

curran:
I meant that in the case that you did not delete the first or last tag, but instead you deleted one in the middle of the list, the ideal behavior would be that the sidebar would become blank, as it does when you delete the first or the last tag. However, I couldn't figure out how to make this happen. So, this fix does fix the bug as described, but there is still room for improvement.

In summary:

 * Deleting either the first or the last tag causes the sidebar to become blank, which is the correct behavior and fixes the original bug.
 * Deleting a tag in the middle of the list causes the next tag in the list to be selected, and the sidebar display changes to reflect details of the newly selected tag. This behavior is acceptable, but not ideal. Ideally, deleting a tag from the middle of the list should have the same behavior as deleting either the first or the last - no tag should be selected, and the sidebar should become blank.

kevin-alpine:
@curran: I figured out an additional change that resolves the issues you mentioned above.  Please take a look, I think between the two of us we definitely have DEV-11057 fixed, now.  

I think if you approve, we should go ahead and merge this branch in: LGTM with my additional commit.

curran:
Looks great!

PR #131, opened 2015-08-17
fix to allow only admins to download log files

atul-alpine:
@prakash-alpine please let me know if this is correct or do I need to add migrations in ChorusClass, Operations and Permissions tables.


prakash-alpine:
LGTM.

prakash-alpine:
Verified. LGTM.

PR #132, opened 2015-08-17
Fixed DEV-12159 via CSS

curran:
I took a stab at solving this by adding a left margin to ".content_header h1". I'm not sure if this is the right solution, but it seems to work well and not interfere with anything else.


curran:
This is not a comprehensive solutions, some things are still flush to the left of the page (e.g. the briefcase on the Workspace overview page).

mmilano:
@curran this problem has actually been in chorus across most pages for nearly a year now.
eg: DEV-9577.

on the backlog until i can can have time to really address this.

it is of course possible to address this on a bit by bit basis, 
although that really isnt ideal.

also, if you do make changes, recommend not using hard coded values (like here)
but using a named variable value defined in the _common.scss file
so that it is possible to track and to modify all at once.

right now there is $page-margin-left and $page-margin-right,
but those are @ 25px, and you probably want 10px in the short term for consistently with other pieces of existing pages.

curran:
Indeed, I totally hear what you are saying. Perhaps it's better not to merge this change, and rather wait until you can properly address the whole issue. I just took a quick stab at it to see if this approach would work.

I suppose the full solution would involve establishing a consistently classed hierarchy of containers across all the pages.

mmilano:
ive gone and addressed this with some css refactoring to /content_header and fixing a bleed-over of some worklet css. 
@curran you can delete this branch and pull at your convenience.

curran:
Ok, great! Closing this PR.

PR #133, opened 2015-08-17
DEV-9045 - Fix permissions on multi-job select sidebar

curran:
This change fixes a bug where if am a non-admin, non-member, non owner of a workspace, the right hand side panel is enabled with actions when I select multiple jobs, but should not be.

This fix adds a check for the permissions of the current user against the permissions of all selected jobs. The list of actions in the sidebar (Disable, Enable, Delete) is only displayed if the user has permissions to update ALL of the currently selected jobs.

### DIFF
--- a/app/assets/javascripts/views/multiple_selection_sidebar_menu.js
+++ b/app/assets/javascripts/views/multiple_selection_sidebar_menu.js
@@ -47,7 +47,16 @@ chorus.views.MultipleSelectionSidebarMenu = chorus.views.Base.include(
 
         additionalContext: function () {
             var actions = _.map(this.actions, this.templateValues);
+
+            // Only show the list of actions in the sidebar (Disable, Enable, Delete)
+            // if the currently logged in user has permissions to
+            // update ALL of the currently selected jobs.
+            var canUpdateSelected = this.selectedModels.every(function (selectedModel){
+                return selectedModel.workspace().canUpdate();
+            });
+
             return {
+                canUpdateSelected: canUpdateSelected,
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/views/multiple_selection_sidebar_menu.js
```diff
@@ -47,7 +47,16 @@ chorus.views.MultipleSelectionSidebarMenu = chorus.views.Base.include(
 
         additionalContext: function () {
             var actions = _.map(this.actions, this.templateValues);
+
+            // Only show the list of actions in the sidebar (Disable, Enable, Delete)
+            // if the currently logged in user has permissions to
+            // update ALL of the currently selected jobs.
+            var canUpdateSelected = this.selectedModels.every(function (selectedModel){
+                return selectedModel.workspace().canUpdate();
```
kevin-alpine: Ok, I found a problem here -- the problem is that `multiple_selection_sidebar_menu.js` is an abstracted class that is used with:

data_source_index_page.js, hdfs_entry_index_page.js, schema_dataset_index_page.js, workspace_dataset_index_page.js, jobs_index_page.js, kaggle_user_index_page.js, search_index_page.js, user_index_page.js, workfile_index_page.js, workspace_index_page.js ...

So, we can't have this reference to `selectedModel.workspace()`, it is not abstract enough.

curran:
Oh wow, thanks for catching that. I will continue working on this today.

curran:
LGTM, ready for merge into master.

PR #136, opened 2015-08-18
Fix for [DEV-12309]  home page widget "site snapshot" takes very long…

prakash-alpine:
Fix for [DEV-12309]  home page widget "site snapshot" takes very log to render and has no loading indicator.


prakash-alpine:
Self LGTM a small change.

PR #138, opened 2015-08-18
DEV-12178 #comment added place holder for thrift help text

atul-alpine:
@prakash-alpine @kevin-alpine  Added the placeholder for thrift help message. Please update the message and check. I didnt had the correct message.Check https://alpine.atlassian.net/browse/DEV-12178


kevin-alpine:
I @atul-alpine -- I understand what's needed.  I pushed up a fix and, now I will merge it in.  Thanks for your work here!

PR #142, opened 2015-08-19
12191: make sure all users can view 'Help & Support' dialog

kevin-alpine:
Currently, the "Help & Support" link does not appear at all, if you are not an admin.  This doesn't seem ideal.  Perhaps it would be better if just the "Download Logs" section doesn't appear, if you are not an admin?

https://alpine.atlassian.net/browse/DEV-12191

@mmilano please review and merge if you agree.


mmilano:
yes, preferable.

PR #141, opened 2015-08-19
Disabled link for Hadoop Hive data sources in Activity feed

curran:
Fixes DEV-12192

PR #143, opened 2015-08-19
12192: allow showUrlTemplate to be a function which return null

kevin-alpine:
Hi @curran,

I reviewed #141 -- it is an ok fix, but I don't like it because it puts a reference is 'isHdfsHive' into `activity_presenters.js` ... which is a file which deals with displaying the activity stream for all model objects ... so for example, it will call `user.isHdfsHive()`, `worklet.isHdfsHive()` and so on for all the models in the `/javascripts/models` directory.

I noticed how link is created by using `showUrlTemplate` ... `showUrlTemplate` can also be a function and its used that way in several files, for example `alpineWorkfile`, `hdfsEntry` and so forth.  What I do here is to enhance the `chorus.Mixins.Urls.showUrl()` function to allow the `showUrlTemplate` function to return null, and in this case then, `showLink` doesn't return a link at all.

I like this better because, we put new abstract functionality into the abstract bits, and then modify `models/hdfs_data_source.js` specifically to use the new abstractly defined functionality.

KT

### DIFF
--- a/app/assets/javascripts/mixins/urls.js
+++ b/app/assets/javascripts/mixins/urls.js
@@ -9,9 +9,15 @@ chorus.Mixins.Urls = {
         }
 
         var template = _.isFunction(this.showUrlTemplate) ? this.showUrlTemplate.apply(this, arguments) : this.showUrlTemplate;
-        var prefix = "#/";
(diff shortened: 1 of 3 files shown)


curran:
Nice work! Indeed, this is a cleaner solution. Thank you for this.

kevin-alpine:
Let's merge 'er on in!

curran:
Ay!

PR #144, opened 2015-08-19
12441 Touchpoint 'Save and Close' does not save changed image

kevin-alpine:
When `this.closePage();` is called, it must deconstruct the view objects, and it cancels any `workletSaved` callbacks.  So, we have to move the `worklet.saveFiles();` call into the actual 'save' machinery.

This also fixes another bug where, if you upload a changed image, and then switch tabs, the image would be broken.  This reloads the image.


jleealpine:
Thanks Kevin, these changes look good, feel free to merge this into master

kevin-alpine:
Thanks @jleealpine: merging it in.

PR #145, opened 2015-08-19
Fix for [DEV-12064] Error message not appearing when attempting to vi…

prakash-alpine:
Fix for [DEV-12064] Error message not appearing when attempting to view a non-shared DB data source after upgrade.


kevin-alpine:
Prakash and I reviewed this for about an hour just now -- Prakash convinced me this is correct.  :)  It LGTM!

PR #146, opened 2015-08-20
DEV-9258 Fix to add Milestone name in activity for deleted milestones

atul-alpine:
@prakash-alpine @kevin-alpine Please review these changes

### DIFF
--- a/app/assets/javascripts/models/activity.js
+++ b/app/assets/javascripts/models/activity.js
@@ -253,6 +253,10 @@
             return this.get("commitMessage");
         },
 
+        getMilestoneName: function() {
+            return this.get("milestoneName");
+        },
+
         isUserGenerated: function () {
             return this.isNote() || this.isInsight() || this.isSubComment();
         },
--- a/app/assets/javascripts/presenters/activity_presenters.js
+++ b/app/assets/javascripts/presenters/activity_presenters.js
@@ -65,6 +65,10 @@
             return this.model.hasCommitMessage();
         },
 
+        getMilestoneName: function() {
+            return this.model.getMilestoneName();
+        },
+
         isWorkfileImage: function() {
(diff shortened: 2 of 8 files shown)


app/controllers/events_controller.rb
```diff
@@ -23,9 +23,12 @@ def index
     #@options =  { :workspace => workspace , :user => current_user, :rendering_activities => true, :show_latest_comments => false}
     #TODO Scope Filter results for current user's scope
     events = Events::Base.filter_by_scope(current_user, events) if current_user_in_scope?
+    events = Activity.remove_events_with_deleted_milestones(events)
     present paginate(events), :presenter_options => {:activity_stream => true, :succinct => true,
                                                       :workfile_as_latest_version => true, :cached => true, :namespace => 'activities'}
```
prakash-alpine: @atul-alpine This does not seem like the right place to filter out events with nil milestone names. This is only needed for existing customers who may have deleted milestones. For new customers this will be unnecessary. I am wondering it it will be better to apply the filter on the client side while parsing JSON data.

CC: @kevin-alpine  Can you chime in?

kevin-alpine:
I agree with @prakash-alpine, the javascript should be able to do this itself without needing to filter the API at the Rails level.  I will review it further after I have made progress on a much higher priority issue: https://alpine.atlassian.net/browse/DEV-12054

About the timing of this: this is a risky patch to merge into 5.6 at this point.  The bulk of https://alpine.atlassian.net/browse/DEV-9258 has been present in master for weeks.  This is icing on top of a 95% finished feature.  I question why we need to merge this into master now.

atul-alpine:
@prakash-alpine @kevin-alpine thanks for the feedback. I will fix this at the backbone side.

alpinebuilder:

PR #147, opened 2015-08-21
Eliminate gray line in empty jobs sidebar on multi-select

curran:
This pull request fixes the gray line issue pointed out by @maniiyer in https://alpine.atlassian.net/browse/DEV-9045

When multiple jobs are selected and user does not have permissions to update them, they were not shown but there was a gray line on the right side panel that should not have been there. I modified the solution such that when no actions are listed, the sidebar container is hidden (which removes the gray line) rather than having an empty template rendered into it.

Also added missing "return" statement in app/assets/javascripts/models/job.js canUpdate function. Verified locally that an admin can select multiple jobs and sees the actions, but a user without update permissions does not see the actions.

### DIFF
--- a/app/assets/javascripts/views/multiple_selection_sidebar_menu.js
+++ b/app/assets/javascripts/views/multiple_selection_sidebar_menu.js
@@ -29,7 +29,23 @@ chorus.views.MultipleSelectionSidebarMenu = chorus.views.Base.include(
         },
 
         revealOnlyMultiActions: function () {
-            $('.multiple_selection').removeClass('hidden');
+
+            // Show the actions list only if the user has permissions
+            // to update all of the selected models.
+            var canUpdateSelected = this.selectedModels.every(function (model) {
+                if (typeof model.canUpdate == 'function') {
+                    return model.canUpdate();
+                } else {
+                    return true;
+                }
+            });
+
+            if(canUpdateSelected){
+                $('.multiple_selection').removeClass('hidden');
+            } else {
+                $('.multiple_selection').addClass('hidden');
+            }
+
             $('#sidebar').find('.primary').addClass('hidden');
         },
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/views/multiple_selection_sidebar_menu.js
```diff
@@ -29,7 +29,23 @@ chorus.views.MultipleSelectionSidebarMenu = chorus.views.Base.include(
...
+                    return model.canUpdate();
+                } else {
+                    return true;
+                }
+            });
+
+            if(canUpdateSelected){
+                $('.multiple_selection').removeClass('hidden');
+            } else {
+                $('.multiple_selection').addClass('hidden');
```
kevin-alpine: Strictly speaking, this else clause ... `$('.multiple_selection').addClass('hidden');` doesn't belong here -- the div is already hidden ... and this method reveals it ... now we're modifying it, to potentially reveal it.  

But, this method never needs to hide it.

kevin-alpine:
Hi @curran,

Besides for the one minor comment, this LGTM!

KT

app/assets/javascripts/views/multiple_selection_sidebar_menu.js
curran: I did consider this, and based on the method name it seems that it should only "reveal", but I wasn't sure if I could safely assume that it was already hidden. How can we know it will always be hidden when this code runs? I didn't see any place where it was initialized to be hidden before this method gets invoked, so I figured I would make this piece of code either add or remove the class, so we know that whether it was hidden or shown before, it will be in the correct state.

curran: Consider, for example, the case that a set of jobs are multi-selected that the user has access to (in which case the actions will be shown), and then after that, the user adds a job to the selection that he/she does not have access to. This should cause the sidebar to change from having the "hidden" class to not having the "hidden" class, and I think this case is not covered if the code were to only remove the hidden class.

curran: Actually, I now realize that this case can never occur, because permissions on jobs defers to permissions on the containing workspace, so at any given instance, a user will either have access to all jobs or no jobs. So I think I will make the change you suggest.

curran:
Thanks @kevin-alpine , I made the change you pointed out and verified that everything still works as expected. Will now merge.

PR #153, opened 2015-08-26
Changed activity presenter to not create links when model URL is null

curran:
This fixes the issue that Hive Hadoop data sources were showing a link (`<a>` tag) with no `href` attribute. This was making the text look normal (black), but when the user hovered over it, it gave the cursor affordance that the text was clickable.

This pull request adds code that checks to see if the URL is defined before creating the link. If the URL is not defined (as in the case of Hive Hadoop data sources), the model name string is returned directly and not wrapped in an `<a>` element.

### DIFF
--- a/app/assets/javascripts/presenters/activity_presenters.js
+++ b/app/assets/javascripts/presenters/activity_presenters.js
@@ -738,7 +738,16 @@
         },
 
         modelLink: function(model) {
-            return Handlebars.helpers.linkTo(model.showUrl(), model.name());
+            var url = model.showUrl();
+            
+            if(url){
+                return Handlebars.helpers.linkTo(url, model.name());
+            } else {
+
+                // If the URL is null, do not create an <a> element,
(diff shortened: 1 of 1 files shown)


mmilano:
@curran how does the title attribute get included and set, then?

kevin-alpine:
This LGTM Curran!  Thanks for your work.

kevin-alpine:
Hi @mmilano -- Curran is changing the behavior of the link that appears in the sidebar ... not the one that appears in the main pane.

alpine chorus

curran:
Indeed, and the link that appears in the global activity feed. This change should not affect any title attribute.

curran:
Thank you Kevin for reviewing, merging.

mmilano:
@curran ah, ok. thats a different link then. thanks.

this will require changes to the test spec's btw. you should also change those.

curran:
The tests pass right now, but I suppose I should add a test for the new case I introduced.

PR #154, opened 2015-08-26
5.6 related test fixes

kevin-alpine:
There are ~25 failing specs in master, left over from the 5.6 push.  There are undoubtedly failing Jasmine tests as well.

Working through them will probably reveal some bugs, but they are probably mostly a result of the tests having fallen behind the functional code.  

Let's collaborate on the fixes in this PR.  

I'm not sure if we can merge this in before 5.6 is released, or if we should wait till afterwards?  I think we can discuss with @nkapinos and @sasher.

@AndrewLngdn @curran @jleealpine @midnighteuler @mmilano

### DIFF
--- a/spec/controllers/alpine/credentials_controller_spec.rb
+++ b/spec/controllers/alpine/credentials_controller_spec.rb
@@ -41,9 +41,9 @@ describe Alpine::CredentialsController do
       context 'when the user does not have an account' do
         let(:user) { users(:not_a_member) }
 
-        it 'returns 404 when the user does not have an account' do
+        it 'returns 403 Forbidden when the user does not have an account' do
           get :show, :data_source_id => data_source.id
-          response.should be_not_found
+          response.should be_forbidden
         end
       end
 
--- a/spec/controllers/workfile_results_controller_spec.rb
+++ b/spec/controllers/workfile_results_controller_spec.rb
@@ -6,7 +6,8 @@ describe WorkfileResultsController do
     let(:workfile) { workfiles(:alpine_flow) }
     let(:params) do
       {
-        :workfile_id => workfile.id
+        :workfile_id => workfile.id,
+        :results_written => 'true'
       }
     end
 
@@ -29,6 +30,7 @@ describe WorkfileResultsController do
       let(:params) do
         {
           :workfile_id => workfile.id,
+          :results_written => 'true',
           :result_id => "123"
         }
       end
--- a/spec/lib/task_helpers/bulk_data_spec.rb
+++ b/spec/lib/task_helpers/bulk_data_spec.rb
@@ -25,7 +25,7 @@ describe BulkData do
     before do
       BulkData.create_fake_users(file_name, 2)
       CSV.open(file_name, 'a') do |file|
-        file << %w(Charlie Chambers chch t chch@somewhere.com ThisGuy Nowhere hahaha)
+        file << %w(Charlie Chambers chch true chch@somewhere.com ThisGuy Nowhere hahaha)
       end
     end
 
--- a/spec/lib/task_helpers/bulk_data_spec.rb
+++ b/spec/lib/task_helpers/bulk_data_spec.rb
@@ -25,7 +25,7 @@ describe BulkData do
     before do
       BulkData.create_fake_users(file_name, 2)
       CSV.open(file_name, 'a') do |file|
-        file << %w(Charlie Chambers chch true chch@somewhere.com ThisGuy Nowhere hahaha)
+        file << %w(Charlie Chambers chch t chch@somewhere.com ThisGuy Nowhere hahaha)
       end
     end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -140,7 +140,7 @@ class User < ActiveRecord::Base
     app_manager_role = Role.find_by_name("ApplicationManager")
     site_admin_role = Role.find_by_name("SiteAdministrator")
 
-    if value == true || value == "true"
+    if ActiveRecord::ConnectionAdapters::Column.value_to_boolean(value)
 
       admin_role.users << self unless admin_role.users.include? self
       app_manager_role.users << self unless app_manager_role.users.include? self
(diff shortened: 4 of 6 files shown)


spec/controllers/alpine/credentials_controller_spec.rb
```diff
@@ -41,9 +41,9 @@
       context 'when the user does not have an account' do
         let(:user) { users(:not_a_member) }
 
-        it 'returns 404 when the user does not have an account' do
+        it 'returns 403 Forbidden when the user does not have an account' do
           get :show, :data_source_id => data_source.id
-          response.should be_not_found
+          response.should be_forbidden
```
kevin-alpine: This is to fixing the failing spec:
    
```
    1) Alpine::CredentialsController#show when alpine is enabled when the user does not have an account returns 404 when the user does not have an account
          Failure/Error: response.should be_not_found
            expected not_found? to return true, got false
          # ./spec/controllers/alpine/credentials_controller_spec.rb:46:in `(root)'
```

That came from Prakash's commit:
https://github.com/Chorus/chorus/commit/7a19f409b470dbf77169d1743765dcc21b33052e#diff-97d8f17d61cfd2136dbca2316389ac67R135

Prakash made the commit to fix:
https://alpine.atlassian.net/browse/DEV-12064

Eventually this goes back to @AndrewLngdn's "Roles and Permission" changes -- Prakash and I discussed the best fix to make, in the Jira ticket, and we had [a question for Andrew](https://alpine.atlassian.net/browse/DEV-12064?focusedCommentId=60670&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-60670)

@AndrewLngdn: the spec is fixed by this change, but based on all that background you may want to revisit Prakash's fix?  I will leave it up to you.

spec/controllers/workfile_results_controller_spec.rb
```diff
@@ -29,6 +30,7 @@
       let(:params) do
         {
           :workfile_id => workfile.id,
+          :results_written => 'true',
```
kevin-alpine: These two changes are to fix these failing specs:

```
  2) WorkfileResultsController#create should make a WorkfileResult event
     Failure/Error: expect {
       count should have been changed by 1, but was changed by 0
     # ./spec/controllers/workfile_results_controller_spec.rb:18:in `(root)'

  3) WorkfileResultsController#create should return a 201
     Failure/Error: response.status.should == 201
       expected: 201
            got: 200 (using ==)
     # ./spec/controllers/workfile_results_controller_spec.rb:25:in `(root)'

  4) WorkfileResultsController#create when there is a result_id creates an attachment for the WorkfileResult event
     Failure/Error: expect {
       count should have been changed by 1, but was changed by 0
     # ./spec/controllers/workfile_results_controller_spec.rb:38:in `(root)'
     # ./spec/controllers/workfile_results_controller_spec.rb:37:in `(root)'
```

Changes in this file, are very similar to fixes made by @jleealpine: https://github.com/Chorus/chorus/commit/80f13dd2a9f95c1a9b4e6867d2a42bbcb3cb01cd

midnighteuler:
@kevin-alpine, @AndrewLngdn:
The 5 commits above should correct the 23 jasmine test failures.

spec/lib/task_helpers/bulk_data_spec.rb
```diff
@@ -25,7 +25,7 @@
     before do
       BulkData.create_fake_users(file_name, 2)
       CSV.open(file_name, 'a') do |file|
-        file << %w(Charlie Chambers chch t chch@somewhere.com ThisGuy Nowhere hahaha)
+        file << %w(Charlie Chambers chch true chch@somewhere.com ThisGuy Nowhere hahaha)
```
kevin-alpine: Due to @AndrewLngdn's "Roles & Groups" related changes, there is a subtle change in the accepted CSV file format for "bulk uploading" users.  Before, an admin could be indicated by a "t" character.  Now it must be spelled out as "true".  You can see the code related to this, here: https://github.com/Chorus/chorus/blob/71497ece7530e38929a8f95858890a09431501a8/app/models/user.rb#L144

spec/controllers/alpine/credentials_controller_spec.rb
AndrewLngdn: LGTM, it isn't super important which error we through as long as it triggers the correct dialogue box

spec/lib/task_helpers/bulk_data_spec.rb
AndrewLngdn: Should we update the User model to accept 't', or is this something that is unlikely to re-occur?

kevin-alpine: I think, what we should do is figure out if anybody ever uses a "t" in their CSV files.  Otherwise, maybe we could hook into how ActiveRecord determines "truthiness" ... I don't think we should hardcode all the possibilities into that big OR statement in `User.rb`.  I thought of fixing the bug in that way but it didn't seem right, it could go on forever.

AndrewLngdn: I agree, I think we should try to figure out how AR determines truthiness. If we look at the [old method](https://github.com/Chorus/chorus/blob/f257f891b5542c42bbe5e039eb29855b000c7a39/app/models/user.rb#L108), write_attribute used to be fine with `true`, `'true'`, and `'t'`. When I switched it to an if-statement, we lost that flexibility. 

If the CSV is for bulk-uploading users, it's possible that it came from a DB dump that stored true as t. I'm not sure where the test case came from, but it's possible that it's a real-world example of what the CSV data could look like

kevin-alpine: There's really no urgency here.  If it's not too much trouble, perhaps you could ask Sara or one of the support people for an example file that's used with this code?

For the other route, here's a StackOverflow question that's relevant:
http://stackoverflow.com/questions/22372593/how-are-values-cast-when-assigning-to-a-boolean-field-in-rails-activerecord

Note this quote: "This is done in the bowels of ActiveRecord: specifically with `ActiveRecord::ConnectionAdapters::Column.value_to_boolean`" ... you could experiment with using that instead of your string comparison stuff.  This way the behavior of the CSV upload wouldn't change.

kevin-alpine:
Andrew and I discussed this and, we will merge this in.  It's very low risk, and we don't want to sit on it any more.

PR #159, opened 2015-09-08
DEV-12704 update user image contoller to allow admins to change profi…

AndrewLngdn:
…le pictures

### DIFF
--- a/app/controllers/user_images_controller.rb
+++ b/app/controllers/user_images_controller.rb
@@ -1,5 +1,4 @@
 class UserImagesController < ImagesController
-  before_filter :require_referenced_user, :only => :create
   protected
 
   def load_entity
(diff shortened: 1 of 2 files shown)


alpinebuilder:

AndrewLngdn:
Hey Kevin, this one should be quick and merge-able

app/controllers/user_images_controller.rb
```diff
@@ -8,8 +7,7 @@ def load_entity
   end
 
   def authorize_create!
-    # Remove this method when 'authorize_create!' is taken out of
-    # ImagesController
+    Authority.authorize! :update, @user, current_user, { :or => :current_user_is_referenced_user }
```
kevin-alpine: Ok I loaded this PR up in my dev instance and tested it out.

It works:
* admins can change anyone's profile image
* nonadmins can only change their own

... one comment, I tried removing this bit, and, the behavior stayed exactly the same, is it necessary?

`{ :or => :current_user_is_referenced_user }`

kevin-alpine: I even hacked Backbone to allow my non-admin user to attempt to change someone else's image, and my attempt at hax0ring was rejected, correctly!

AndrewLngdn: I'll look into this in 5.7, good find!

PR #160, opened 2015-09-09
DEV-12621: check to see if migrate permissions has been run before starting webserver

AndrewLngdn:
The migration check needs to run after Rails initialization, so I saw two choices: 

1. Add a new file under a new config/initializer that performs the proper check
2. Add the code to the application.rb file, and call the function after_initialize

1. Would decouple the code from the application.rb file, but it would not be immediately obvious when/where the code was being called for future developers
2. The code is fairly obvious in the initialization process, but it clutters application.rb a little bit

@kevin-alpine @midnighteuler @prakash-alpine 

Ideas?


alpinebuilder:

prakash-alpine:
I just tried out the solution.  It displays the error on console:

Error: rake db:seed_permissions must be run before rake db:migrate_permissions

However, it then starts the webserver as normal. This is not the ideal behavior. On error condition, we somehow need to prevent the rails process from starting..

app/mixins/permissioner.rb
```diff
@@ -18,6 +18,24 @@ module Permissioner
...
+    end
+
+    user_co_count = ChorusObject.where(:chorus_class_id => user_class.id).count
+    ws_co_count = ChorusObject.where(:chorus_class_id =>  workspace_class.id).count
+    datasource_co_count = ChorusObject.where(:chorus_class_id => datasource_class.id).count
+
+
+    user_co_count != User.count || ws_co_count != Workspace.count || datasource_co_count != DataSource.count
+  end
+
```
prakash-alpine: @AndrewLngdn   Permissioner is a mixin and any method added to it will get exposed to all ActiveRecord classes that include it.

should_migrate_permissions? is a utility method that is better placed in a library module than a mixin.

prakash-alpine:
@AndrewLngdn I made few changes to simplify the process

1. Added separate check should_seed_permissions? to permissions_utils.rb
2. Added seed_permissions task to "chorus_control.sh migrate" command so it does not have to be run
    separately

I have tested all possible scenarios from command line. But you should test and verify that I have not regressed something unintentionally.  Also check with Hao to make sure the change to chorus_control.sh is OK.

Ping me on skype if you have any questions.

PR #165, opened 2015-09-18
Renables the better version of .filter_by_scope and...

AndrewLngdn:
introduces the better version of the permissions migrator. Also adds tests for every scoped endpoint, and fixes a bunch of broken permissions. 

PermissionsMigrator and filter_by_scope are the biggest things here, the rest is just cleaning stuff up. Both are tested. 

Filter_by_scope looks complicated, and it's harder to read than the old method. However, it performs the minimal amount of SQL queries necessary to get the scoped_object_ids in this architecture, without resorting to pure, nested SQL queries (it's basically the nested query unrolled into AR functions). This is an area where performance is important. Prakash's function performed (user.groups.count * ( objects.count + 1) ) number of queries, which could get into the thousands(!) on large endpoints. The new function performs a constant amount of queries and performs the filtering at the end. We could gain a little speed here by checking set inclusion instead of using Array#include? but I'm not sure if the set creation overhead is worth the benefits. It's pretty fast, but we should test it on ChorusAdmin's notifications endpoint because that is the largest (although it shouldn't be returning a thousand objects...)

The PermissionsMigrator is a replacement for the migrate_permissions.rb script. It should be fairly straightforward, especially when compared to [the original](https://github.com/Chorus/chorus/blob/master/db/migrate_permissions.rb)

@kevin-alpine @midnighteuler please leave a comment on anything that looks weird, confusing, or wrong. I'd like to merge this in by mid-day tomorrow so we can get the new migrator into a real rails migration.


alpinebuilder:

kevin-alpine:
Hi @AndrewLngdn -- I am reviewing this now, my apologies that it took me a few days to clear the space.  I will do a good job now.

app/mixins/permissioner.rb
```diff
@@ -20,6 +20,7 @@ module Permissioner
 
   # Returns true if current user has assigned scope. False otherwise
   def self.user_in_scope?(user)
+    return true
     if self.is_admin?(user)
       return false
     end
```
kevin-alpine: Hmm, I think this `return true` at the very beginning of this method will break the method, it may be some debugging code that slipped in?

lib/permissions_migrator.rb
```diff
@@ -0,0 +1,88 @@
...
+      ChorusObject.connection.execute "INSERT INTO chorus_objects #{attribute_columns_string} VALUES #{rows.join(", ")}" if rows.any?
+    end
+  end
+
+  def self.create_rows(klass)
+    scope_id = ChorusScope.where(:name => 'application_realm').pluck(:id).first
+    time = Time.now.to_s(:db)
+
+    klass.all.map do |object|
+      chorus_class_id = ChorusClass.where(:name => object.class.name).pluck(:id).first
```
kevin-alpine: When I ran this migration on my dev instance, I got an exception complaining:

     org.postgresql.util.PSQLException: ERROR: syntax error at or near ","

... after digging into it, I found that sometimes this `chorus_class_id` is null, and when this happens, the `attribute_string` looks like this: `(, 470, NULL, '2015-09-24 17:03:07', '2015-09-24 17:03:07', 2)`

So, I edited this bit to look like this:

    rows = klass.all.map { |object|
      # Why pluck.first? Why not .first.try(:id)
      chorus_class_id = ChorusClass.where(:name => object.class.name).first.try(:id)
      if chorus_class_id
        create_attribute_string(object, chorus_class_id, time, scope_id)
      end
    }.compact

lib/permissions_migrator.rb
```diff
@@ -0,0 +1,88 @@
...
+  def self.insert_chorus_object_rows
+    CLASSES.each do |klass|
+      rows = create_rows(klass)
+
+      ChorusObject.connection.execute "INSERT INTO chorus_objects #{attribute_columns_string} VALUES #{rows.join(", ")}" if rows.any?
+    end
+  end
+
+  def self.create_rows(klass)
+    scope_id = ChorusScope.where(:name => 'application_realm').pluck(:id).first
```
kevin-alpine: I have another comment about this, I think this is pretty inefficient because, it's going to iterate over all the ChorusScopes, and make an array out of the IDs ... then return the first one.  Maybe instead it could be:

`scope_id = ChorusScope.where(:name => 'application_realm').first.try(:id)`

lib/permissions_migrator.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  def self.create_attribte_string(object, chorus_class_id, time, scope_id)
+
+    if object.respond_to? :owner_id
+      object_owner = object.owner_id
+    else
+      object_owner = "NULL"
+    end
+
+    "(#{chorus_class_id}, #{object.id}, #{object_owner}, '#{time}', '#{time}', #{scope_id})"
```
kevin-alpine: If you rename `time` to `created_at` and `updated_at` (or something halfway between like `modified_at`) it will make this slightly more familiar and easier to read.

lib/permissions_migrator.rb
```diff
@@ -0,0 +1,88 @@
...
+    Workspace.all.each do |ws|
+      ws.add_user_to_object_role(ws.owner, owner_role)
+
+      ws.members.each do |member|
+        ws.add_user_to_object_role(member, project_manager_role)
+      end
+    end
+  end
+
+  def self.create_attribte_string(object, chorus_class_id, time, scope_id)
```
kevin-alpine: There are 2 places where "attribute" is misspelled as `attribte` ...

app/mixins/permissioner.rb
```diff
@@ -133,24 +134,15 @@ module ClassMethods
 
     # Given an collection of objects, returns a collection filterd by user's scope. Removes objects that are not in user's current scope.
```
kevin-alpine: `filterd` -> `filtered`

app/mixins/permissioner.rb
AndrewLngdn: This is how it should be, the method is going to be short-circuited until 5.8

lib/permissions_migrator.rb
AndrewLngdn: This error `org.postgresql.util.PSQLException: ERROR: syntax error at or near ","`
indicates that seed_permissions.rb didn't create all the necessary classes. That script has changed so you should re-run rake db:seed_permissions. 

Bypassing `create_attribute_string(object, chorus_class_id, time, scope_id)` will cause incorrect data in the database. The replacement code will not work correctly as the task should fail. It will only fail on dev machines with incorrect states

lib/permissions_migrator.rb
AndrewLngdn: Ya that looks better, 
`scope_id = ChorusScope.where(:name => 'application_realm').first.try(:id)` uses a limit. Although the replaced query is  <1ms and pluck doesn't create Rails models anyway so I don't think performance is a problem here.

lib/permissions_migrator.rb
AndrewLngdn: Good call, did you change this or should I?

app/mixins/permissioner.rb
kevin-alpine: Ahh -- ok, let's leave it commented out in that case, perhaps with a TODO, it is less confusing that way.

lib/permissions_migrator.rb
kevin-alpine: What we could do is, something like this:

```
unless chorus_class_id
  raise "seed_permissions.rb didn't create all the necessary classes. That script has changed so you should re-run rake db:seed_permissions.
end
```

(I just copypasted your text, might edit a bit for max helpfulness.)

lib/permissions_migrator.rb
```diff
@@ -0,0 +1,88 @@
...
+      Job,
+      Milestone,
+      Tag,
+      Upload,
+      Import,
+      Notification,
+      CsvFile,
+      Database,
+      Dataset
+  ]
```
kevin-alpine: What I worry about with this is, this list will fall out of date as we add, rename, or delete models.  The RGP code introduces a significant new thing that we must keep in sync ... we must keep the database state in sync with the list of models.  I don't have a solution to this but, it is something to be aware of and perhaps we could make the design more elegant in time.

I agree with you keeping the PermissionsMigrator out of the db/migrations directory, because this file will likely need to be modified for each release to catch the RGP database state up with any changes made in the models ... it will become significantly more complex/sophisticated over time.

lib/permissions_migrator.rb
AndrewLngdn: good idea, added that line

kevin-alpine: Hmm ok I re-ran `rake db:seed_permissions` but, I still get the error ...

kevin-alpine: So, the model that has no ChorusClass is: `Events::CredentialsInvalid` -- hmm ... what's the plan for all these STI models?  I don't see them listed in the `rake db:seed_permissions` code ... it may be something I have borked up in my environment though?

kevin-alpine:
Hi @AndrewLngdn: I cleaned up the slight tweaks that I came up with yesterday and put them in another PR #168 which is to merge into this PR.  So, if you have not responded to my comments yet, you can just review and merge in that branch, it will make a lot of my comments above, obsolete.

KT

PR #168, opened 2015-09-25
Tweaks for rgp-scopes branch

kevin-alpine:
Committing minor edits that I came up while while reviewing Andrew's branch rgp-scopes.

#roles_groups_permissions #code_cleanup #tweaks


alpinebuilder:

AndrewLngdn:
Looks good, although I'm not sure what happens in
permissions_migrator.rb:40 when there is a chorus class but no objects in
that table. Does spec/lib/permissions_migrator_spec.rb still pass? I don't
know if I have a test case for that. It may generate an improper sql
statement as I believe that was the reason for the if statement

PR #173, opened 2015-10-02
Making VisChiasm work (REVIEW 4th, MERGE 1st)

kevin-alpine:
In this PR I add two APIs that will return the data that Curran needs ... you can access them at these two URLs:
http://localhost:3000/datasets/62270/chiasm_api_datasets/show_data
http://localhost:3000/datasets/62270/chiasm_api_datasets/show_column_data
_... where 62270 is a dataset ID._

I also catch the VisChiasm component up to where the integration that @midnighteuler & @curran  were working on in master, is at ... (I copied over the work that was in master, into the VisChiasm component.)

Now that we have sorted out the files that are concerned with Visualization, we can see that a lot of the Backbone code is broken right now in this component.  I went through and added TODOs and comments asking @curran  & @midnighteuler to adapt the code to be useful again or to delete it.

In the guts of the Rails sever-side query generating code, I deleted a lot of the files which had to do with "presentation specific" database queries ... and just replaced this with a 'random_sample' query method ... for @mthyen to fill in.  (I have removed all the "models" and "presenters" and it's all done very simply (for now) inline in the controller.)

By looking through this PR we can now see:
* Where to put the random_sampling database-specific queries.
* What backbone files are involved with visualizations and how they're broken now ...

Additionally since we still have the VisLegacy component which will not be altered, we can delete and munge this code as much as we like and, if it all goes south, we can just switch back to using the VisLegacy component and take time we need to make VisChiasm good.

I will discuss this work with @curran, @midnighteuler, and @mthyen tomorrow when I am back in the office.


alpinebuilder:

components/vis_chiasm/app/assets/javascripts/mixins/dataset_content_details_visualizations.js
```diff
@@ -11,50 +11,188 @@ chorus.Mixins.DatasetContentDetailsVisualizations = {
```
kevin-alpine: @midnighteuler: ... these are the customizations you made to `dataset_content_details.js` -- but, I broke the visualization specific stuff out into this mixin, so, we can change this mixin, but the VisLegacy version of it, will remain the same.  If you need to edit the actual `dataset_content_details.js` be sure to adapt VisLegacy to still work with any changes you make ...

components/vis_chiasm/app/assets/javascripts/templates/dataset_content_details.hbs
```diff
@@ -1,5 +1,40 @@
 <!-- KT TODO: This doesn't belong in the component but, it's too hard to pull it apart, right now. -->
```
kevin-alpine: This file has to do with a lot more than just visualizations but I didn't have time to figure out how to break it into components.  I think it should be fairly easy with handlebar partials.  Maybe we can chat about this tomorrow @midnighteuler?

components/vis_chiasm/app/controllers/chiasm_api_datasets_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    @sql_generator = @connection.visualization_sql_generator
+
+    # TODO Curran / Michael Thyen -- passing in :foo so we have the option to pass multiple named options,
+    # otherwise we can clean it up (just pass dataset)
+    row_sql = @sql_generator.random_sampling_sql(:dataset => dataset, :foo => 'bar')
+
+    # TODO Curran / Mike Souza see VisLegacy -> VisualizationsController#destroy -- the query must be cancellable for
+    # a reason, we will probably need to wire up something equivalent, passing the 'check_id' along
+    # as a key for the cancelling of query.
+    result = CancelableQuery.new(@connection, check_id, current_user).execute(row_sql)
```
kevin-alpine: @curran, @midnighteuler, @mthyen note this method: it's the heart of the 'random_sampling' api ... I put some notes inline for you guys.

components/vis_chiasm/app/controllers/visualizations_controller.rb
```diff
@@ -1,19 +0,0 @@
...
-    visualization = Visualization.build(dataset, params[:chart_task])
-    visualization.fetch!(authorized_account(dataset), params[:chart_task][:check_id])
-    present visualization
-  end
-
-  def destroy
-    dataset = Dataset.find(params[:dataset_id])
-    authorized_account(dataset)
-    CancelableQuery.cancel(params[:id], current_user)
-    head :ok
```
kevin-alpine: We need to wire this back up, there's no way to cancel the `CancelableQuery` right now.

components/vis_chiasm/app/models/visualization.rb
```diff
@@ -1,51 +0,0 @@
...
-  UnknownType = Class.new(StandardError)
-
-  def self.build(dataset, attributes)
-    case attributes[:type]
-      when 'frequency' then Visualization::Frequency
-      when 'histogram' then Visualization::Histogram
-      when 'heatmap' then Visualization::Heatmap
-      when 'timeseries' then Visualization::Timeseries
-      when 'boxplot' then Visualization::Boxplot
-      else raise UnknownType, "Unknown visualization: #{attributes[:type]}"
```
kevin-alpine: This has to do with massaging the data server-side into the right form for different chart types.  This is now done on the client -- so, all this goes away.

components/vis_chiasm/app/services/visualization/hive2_sql.rb
```diff
@@ -5,78 +5,8 @@ def self.extend_object(obj)
...
-      #q
-    end
-
-    def timeseries_row_sql(o)
-      raise NotImplemented
-    end
-
-    def boxplot_row_sql(o)
+    def random_sampling_sql(o)
+      # Michael Thyen TODO -- fill in
```
kevin-alpine: @mthyen: note, the database-specific random sampling SQL goes into these files ...

components/vis_chiasm/app/services/visualization/sql_generator.rb
```diff
@@ -14,169 +14,51 @@ def initialize(params={})
...
+
+      dataset, foo = fetch_opts(o, :dataset, :foo)
+
+      columns = dataset.column_names.sort.join(',')
+      randomFunctionNoSeed = "random()"
+      sourceTable = dataset.scoped_name
+      randomValue = 0.1
+      limit = 10
+
+      "SELECT #{columns} FROM (SELECT #{columns}, #{randomFunctionNoSeed} AS rand_order FROM #{sourceTable}) alpfoo WHERE alpfoo.rand_order <= #{randomValue} LIMIT #{limit}"
```
kevin-alpine: @mthyen: this is the base class for the SQL generating -- if it's not overridden in a database specific subclass, this is what gets executed.  The file `services/jdbc_overrides.rb` is what instantiates the database specific subclasses ... based on the `db_url`.

components/vis_chiasm/app/services/visualization/sql_generator.rb
```diff
@@ -14,169 +14,51 @@ def initialize(params={})
...
-      ) AS total
-        FROM (#{ntiles_for_each_bucket}) AS ntilesForEachBin
-        ORDER BY total desc, "#{category}", ntile #{limits[:limit]};
-      SQL
-
-      ntiles_for_each_bin_with_total
-    end
-
-    def heatmap_min_max_sql(o)
-      dataset, x_axis, y_axis = fetch_opts(o, :dataset, :x_axis, :y_axis)
```
kevin-alpine: I love deleting things!

components/vis_chiasm/config/initializers/assets.rb
```diff
@@ -0,0 +1 @@
+Rails.application.config.assets.precompile += ['vis_chiasm_vendor.js']
```
kevin-alpine: @midnighteuler @AndrewLngdn: there's a classic gotcha here, it works in development, but the asset file does not show up on production -- if that happens, it's because there's something wrong with this line.

components/vis_chiasm/vendor/assets/javascripts/vis_chiasm/visEngineDataLoader.js
```diff
@@ -0,0 +1,44 @@
...
+        dataset_id: Model.None
+    });
+
+    model.when("dataset_id", function(dataset_id) {
+        if (dataset_id !== Model.None) {
+
+            var urls = ['datasets/' + dataset_id + '/chiasm_api_datasets/show_column_data',
+                'datasets/' + dataset_id + '/chiasm_api_datasets/show_data'];
+
+            $.when.apply($, urls.map(function(url) {
```
kevin-alpine: @curran: here is where I clumsily duct-taped this together -- the chart is not quite loading right now, and I am not sure why.

components/vis_chiasm/vendor/assets/javascripts/vis_chiasm/visEngineDataLoader.js
```diff
@@ -0,0 +1,44 @@
...
+                return $.ajax({
+                    url: url,
+                    async: false,
+                    dataType: 'text'
+                });
+            })).done(function() {
+
+                var column_data = eval(arguments[0][0]);
```
kevin-alpine: @curran: this `eval(arguments[0][0])` seems wrong but, I didn't take the time to patch it up.  This is something we should figure out tmr.

components/vis_chiasm/vendor/assets/javascripts/vis_chiasm/visEngineDataLoader.js
curran: This is great, I can take this and run with it. I can take the URL patterns here and adapt this file to use them https://github.com/chiasm-project/chiasm-dsv-dataset/blob/master/index.js

components/vis_chiasm/vendor/assets/javascripts/vis_chiasm/visEngineDataLoader.js
curran: I think we can use JSON.parse here .. Let's discuss.

kevin-alpine:
Regarding the tests: 

* The Rails specs are in pretty good shape.  In the vis_legacy branch they're all behaving normally.  In vis_chiasm we just have to delete the specs to match the models and presenters and such that I deleted.

* There seem to be some Jasmine tests failing in vis_legacy related to the componentization ... I'm not sure why this is happening, for example in vis_legacy if we run http://localhost:8888/?spec=chorus.dialogs.Visualization like so, we get these sorts of failures:

```
chorus.dialogs.Visualization #initialization when the save completes when the task data is valid should draw the chart
No fixture for backbone/databaseColumnSet thrown
Error: spyOn could not find an object to spy upon for drawChart()
Error: spyOn could not find an object to spy upon for drawChart()
    at SpyRegistry.spyOn (http://localhost:8888/__jasmine__/jasmine.js:1887:15)
```

... I'm not sure why it's not finding it, because I see the files are being loaded (which is cool because, we know that the Asset Pipeline is working ok via `spec/javascripts/support/jasmine.yml` 

vis_legacy

... it might come down to a load order thing.  I will be back on ~Oct 21st or 22nd.  So if there's no chance for someone else to investigate this, I think there will be time to me to look at this again during the QA period.

KT

curran:
magicbarchartinchorus

Chiasm is working in Chorus with the [Magic Bar Chart (Browserified)](http://bl.ocks.org/curran/a931688fa8cb65db4079) on top of the API that Kevin created for random sampling on database. The user can change the column (numeric columns only right now, temporarily), switch between horizontal and vertical, and change the number of bins. There are lots of issues to work out, but a solid foundation is now there.

Chiasm blockers for 5.7 that we need to address next week include:

 * creating all the other charts besides Bar Chart
   * Heat Map
   * Line Chart
   * Box Plot
 * solving tick mark labels that overlap with the Y axis label (draggable margins maybe?)
 * solving tick mark labels that overlap with each other on the X axis (change tick format? add UI for specifying tick mark density?)
 * solving the case where there are way too many tick mark labels for categorical columns (show tick labels only on hover?)
 * implementing a random sampling function on SQL based on the number of rows rather than a percentage (Michael T. can potentially help with this)
 * adding UI widgets for all the configurable parameters we want to expose (Mike S. can potentially help with this)
 * getting the image saving to work (hopefully we can hook into the existing infrastructure)
 * integrating filters from the Chorus UI into the visualization computation
 * support changing the column for all types (need to somehow incorporate the schema data returned by the API in the UI code)

PR #174, opened 2015-10-02
OSX ImageMagick: just run `brew install imagemagick`

kevin-alpine:
Jason Lee got an exception when using my statically compiled ImageMagick for OSX.  I think he may be using a different OSX version (not Mavericks). I can't be bothered to create a statically compiled ImageMagick for all the different OSXes -- why?  It's only used for development and the devs can just use brew.  So here, I remove the statically compiled OSX ImageMagick -- and instead raise an exception if a user on OSX tries to start the app and does not have 'identify' (ImageMagick binary) in the path.

#imagemagick #cleanup #brew

### DIFF
--- a/config/initializers/paperclip.rb
+++ b/config/initializers/paperclip.rb
@@ -1,5 +1,19 @@
+def exception_if_no_imagemagick
+  system('identify --version')
+  result=$?
+  exit_code=result.exitstatus
+  if exit_code.eql?(127) # 127 = File Not Found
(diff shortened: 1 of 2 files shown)


alpinebuilder:

jleealpine:
LGTM

kevin-alpine:
Woot!  Thank you @jleealpine -- if anyone has questions about ImageMagick (in development mode on OSX) Jason has a basic idea of what's going on here.

PR #186, opened 2015-10-21
Data source 5.7 fixes 4

AndrewLngdn:
@midnighteuler @jleealpine


alpinebuilder:

db/migrate/20151021053828_delete_duplicate_rows_in_roles_users.rb
```diff
@@ -0,0 +1,16 @@
+class DeleteDuplicateRowsInRolesUsers < ActiveRecord::Migration
```
AndrewLngdn: @kevin-alpine @kevinmtrowbridge  Can this get a good check? It's probably the most potentially destructive parts of the merge.

It should be deleting duplicates in the roles_users table. Since there's no ID on the join table, we have to use this postgres thing ctid. Should group by role_id, user_id, grab the minimum of this ctid, and then delete everything that's not that single row

midnighteuler:
Looks alright to me.

db/migrate/20151021053828_delete_duplicate_rows_in_roles_users.rb
kevin-alpine: Ok, @midnighteuler and I both checked this in different ways and it seems to work fine.

Here is what I did:

_I ran the query like this:
SELECT * FROM roles_users
WHERE ctid NOT IN (
SELECT min(ctid)
FROM roles_users
GROUP BY role_id, user_id);
this shows all the rows it would delete
it returned 613 rows
there are 799 rows in that table
I copied those out into sublime text
I chose "select all" then Edit > Permute Lines > Unique
there were 186 unique rows left
799 - 186 = 613
this is with my dev db from 2 weeks ago
so by that check it works fine_

PR #187, opened 2015-10-21
Increasing the default workfile size limit, to allow for larger model files.

jthompson6:
@jleealpine @kevin-alpine @midnighteuler @codedeft review?

### DIFF
--- a/config/chorus.defaults.properties
+++ b/config/chorus.defaults.properties
@@ -60,7 +60,7 @@ webserver_threads = 40
 
 # File Size Settings
 
-file_sizes_mb.workfiles = 10
+file_sizes_mb.workfiles = 100
 file_sizes_mb.csv_imports = 100
 file_sizes_mb.user_icon = 5
 file_sizes_mb.workspace_icon = 5


alpinebuilder:

codedeft:
I guess this works?

kevin-alpine:
Hi Jenny,

Increasing this, increases the validation size applied to files attached via [paperclip](https://github.com/thoughtbot/paperclip) ... I don't see any reason why it wouldn't work.  It should be QAed though.

Reasons this might be bad:
* Maybe the HTTP connection will timeout before the 100mb file can be uploaded.  But, I don't think Chorus enforces a timeout.  Some platforms I have used like Heroku, do enforce a *25s timeout* ...
* This might cause the host system to run out of disk space sooner, if there are many 100mb attachments.

If it's necessary I suggest merging it in and QAing it thoroughly,
KT

jthompson6:
@nkapinos can we merge?

jthompson6:
LGTM

PR #189, opened 2015-10-23
for Release_5.7, restart indexer every 24h

kevin-alpine:
In [DEV-13301: Chorus indexer process hangs](https://alpine.atlassian.net/browse/DEV-13301) we had to fess up to the fact that the indexer process is buggy and hangs periodically.

We don't  have time to fix this properly in 5.7, but as a stopgap measure, we will restart the indexer every 24 hours.  Restarting ruby processes every 24 hours is a common practice (I know this is done by Heroku, for example) to deal with memory leaks and keep processes running properly.

#indexer #hanging #workaround

### DIFF
--- a/app/services/service_scheduler.rb
+++ b/app/services/service_scheduler.rb
@@ -37,6 +37,11 @@ class ServiceScheduler
       SolrIndexer.SolrQC.enqueue_if_not_queued('SolrIndexer.refresh_external_data')
     end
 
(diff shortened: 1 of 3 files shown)


alpinebuilder:

kevin-alpine:
This PR is to merge into Release_5.7 branch.  @midnighteuler, can you review please?

PR #193, opened 2015-10-24
fix test

haozou:
after deleting the phd 2.0, we need to fix the test also.


alpinebuilder:

haozou:
LGTM

alpinebuilder:
Code not merged because Hao Zou has committed code in the request.

PR #195, opened 2015-10-26
DEV-13359 remove extra admin roles and disable group role check

AndrewLngdn:
5.7 solution to https://alpine.atlassian.net/browse/DEV-13359, see big comment for details 

@jleealpine @kevinmtrowbridge can one of you guys verify this?

### DIFF
--- a/db/migrate/20151026205156_remove_incorrect_admin_roles.rb
+++ b/db/migrate/20151026205156_remove_incorrect_admin_roles.rb
@@ -1,0 +1,17 @@
+class RemoveIncorrectAdminRoles < ActiveRecord::Migration
+  def up
+    admin_role = Role.find_by(:name => "Admin")
+    manager_role = Role.find_by(:name => "ApplicationManager")
+    site_admin_role = Role.find_by(:name => "SiteAdministrator")
+
+    User.all.each do |user|
+      if user.admin? == false && Permissioner.is_admin?(user)
+        user.roles.delete(admin_role, manager_role, site_admin_role)
+      end
+    end
+  end
+
+  def down
+    # not reversible
+  end
+end
--- a/app/helpers/authority.rb
+++ b/app/helpers/authority.rb
@@ -188,9 +188,10 @@ module Authority
   def self.retrieve_roles(user)
     roles = user.roles.to_a
 
-    user.groups.each do |group|
-      roles.concat(group.roles) if group.roles.empty? == false
-    end
+    # See comment on DEV-13359. Disabled for 5.7 release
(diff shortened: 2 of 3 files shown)


alpinebuilder:

db/migrate/20151026205156_remove_incorrect_admin_roles.rb
```diff
@@ -0,0 +1,17 @@
+class RemoveIncorrectAdminRoles < ActiveRecord::Migration
+  def up
+    admin_role = Role.find_by(:name => "Admin")
+    manager_role = Role.find_by(:name => "ApplicationManager")
+    site_admin_role = Role.find_by(:name => "SiteAdministrator")
+
+    User.all.each do |user|
+      if user.admin? == false && Permissioner.is_admin?(user)
```
kevin-alpine: Hi @AndrewLngdn -- I can see this is an important patch.  Should this be `!Permissioner.is_admin?(user)` -- (ie, add NOT) the way it's written now seems backwards.

jleealpine: If I'm reading this right, this is meant to remove users listed as admin under the roles system who are not actually admins according to Chorus, is that right?

kevin-alpine: Ahh I see.  I am reviewing more in depth now.  Apologies for the delay, I was digging into another issue.

jleealpine:
Apart from the comment chain from Kevin and I, the rest of it looks straightforward enough

kevin-alpine:
I dug into this for about 2 hours.  This investment allowed me to begin to understand what the roles code is doing.  I was able to assign a user the admin role, and using this little snippet I could print out the user that had the admin role incorrectly:

```
admin_role = Role.find_by_name 'Admin'
User.all.each do |u|
  if u.admin? == false && Permissioner.is_admin?(u)
    p "User: #{u.username} has admin_role but is not an admin."
  end
end
```

Then, I ran the migration, and ran my snippet again, and this time the message wasn't printed at all.  So it seems that the migration removed the admin role from my user.

I also get what Andrew is saying about commenting out the groups functionality for now.  I agree this is low risk and seems straightforward.

So, we clearly have to fix DEV-13359, and given where we are at in the release cycle, we should merge this in and test.  So LGTM -- I will go ahead and merge this right now.  I will try to understand more tomorrow.

KT

AndrewLngdn:
Thanks for the inquisitiveness Kevin, I should have been more clear in the PR message. Let me know if you have more questions

PR #199, opened 2015-10-27
Hive JDBC controller updates to allow incomplete

AndrewLngdn:
@jleealpine can you please look over this? It's supposed to allow incomplete saving for hive JDBC and also fix a problem where the owner account validation would happen before the account was updated, so the validation would always fail if you had bad credentials before updating

### DIFF
--- a/app/models/data_source.rb
+++ b/app/models/data_source.rb
@@ -226,11 +226,10 @@ class DataSource < ActiveRecord::Base
   end
 
   def validate_owner?
-    self.changed.include?('host')        ||
-    self.changed.include?('port')        ||
-    self.changed.include?('db_name')     ||
-    self.changed.include?('db_username') ||
-    self.changed.include?('db_password')
+    (self.changed.include?('host')        ||
+    self.changed.include?('port')         ||
+    self.changed.include?('db_name'))     &&
+    should_update_account? == false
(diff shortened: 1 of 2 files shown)


alpinebuilder:

app/models/data_source.rb
```diff
@@ -226,11 +226,10 @@ def update_data_source_account_for_owner
...
   def validate_owner?
-    self.changed.include?('host')        ||
-    self.changed.include?('port')        ||
-    self.changed.include?('db_name')     ||
-    self.changed.include?('db_username') ||
-    self.changed.include?('db_password')
+    (self.changed.include?('host')        ||
+    self.changed.include?('port')         ||
+    self.changed.include?('db_name'))     &&
+    should_update_account? == false
```
jleealpine: Does this need to have "db_name" and "db_username" removed from the validate_owner check?  Would it work as long as should_update_account is true?

AndrewLngdn: Yeah it does, because should_update_account has a more "sophisticated" changed check. Owner account stuff in DS is a little messy now that we allow incomplete everything for data sources. Changing the credentials will still validate the owner account, but now it will validate it *after* the new credentials have been passed in instead of before

PR #201, opened 2015-10-30
Componentization 5.7

kevin-alpine:
I manually merged the `Release_5.7` branch into `componentization`.  This was non-trivial but, I only took about 2 days.

I have some more tweaks to do next week before merging into master, but this is a solid first pass.  I hope to merge it into master by the middle of next week.  

I'm eager to get the ball rolling on componentization, so that we will have the most time possible to deal with it during the early stages of 5.8.


alpinebuilder:

kevinmtrowbridge:
This was merged in via the componentization-final branch and, this PR no longer applies.

PR #151, opened 2015-08-24
Roles Groups & Permission changes - DO NOT MERGE

prakash-alpine:
This branch is to be merged with master after 5.6 release branch is cut.

PR #208, opened 2015-11-24
Simple Admin console w/ Licensing Information screen

kevinmtrowbridge:
When present, if a user is an Admin, they see an "Administration" link in the main Chorus left-hand menu.  Currently, it only contains one page, the "Licensing Information" page, which I created based on Michel's UI specs.

Due to being a "component," this is highly isolated from the rest of the app, and easy to turn on or off.

This is based on the Admin console which Atul submitted in https://github.com/Chorus/chorus/pull/151

This PR also extracts the system fonts from the frontend component and into a `web_style` component.  This will be a place for any UI assets shared by multiple components, such as fonts, colors, logos, "branding type things."

admin

chorus_licensing_screen


alpinebuilder:

PR #211, opened 2015-12-02
DO NOT MERGE: Simplify Authorization (post 5.8)

kevinmtrowbridge:
@AndrewLngdn and I will be collaborating on this branch for the *post 5.8* release.

@AndrewLngdn: there are two branches, the `authorization_simpler` branch, and this one, the `authorization_simpler-first-steps` branch.  This way, I can create a PR to merge this (2nd) branch, into the first one.  This way, it's easy to see all the code I'm proposing and to discuss it in the pull request.

This should be a good start.  I'll go through and comment on the code in the PR now.  Let's collaborate closely.


alpinebuilder:

Gemfile
```diff
@@ -1,9 +1,12 @@
 source 'https://rubygems.org'
 
-path 'components' do
-  gem 'admin'
-  gem 'frontend'
-end
+gem 'admin', :path => 'components/admin'
+gem 'api', :path => 'components/api'
+gem 'authorization', :path => 'components/authorization_simpler'
```
kevinmtrowbridge: By listing the paths to all the components explicitly like this, we can swap out the components just based on being in a different directory -- this way we can make the `authorization_simpler` component, use the same ruby namespace of `Authorization` ...

Gemfile.lock
```diff
@@ -75,11 +79,18 @@ PATH
       sass-rails (= 5.0.1)
       sunspot_rails (= 2.1.0)
       will_paginate (= 3.0.7)
-    authorization (1.0.0)
-      activerecord-jdbcpostgresql-adapter
+
+PATH
+  remote: components/authorization_simpler
```
kevinmtrowbridge: Note how the path is explicitly listed here for each component now, whereas before they were all nested under the `components` path ...

components/admin/app/views/admin/dashboard/index.html.erb
```diff
@@ -13,4 +13,8 @@
       </div>
     </div>
   </div>
+
+  <% if defined?(Authorization::Engine) %>
+    <%= render :partial => 'authorization' %>
+  <% end %>
```
kevinmtrowbridge: I created an area in the `Admin` component to expose details of the Authorization -- it looks like this:

chorus admin - welcome

I'm modifying `admin` component, to point at a partial in the `authorization` component -- so, I wrap it in the `if defined?(Authorization::Engine)` block.

components/api/app/controllers/api/users_controller.rb
```diff
@@ -1,8 +1,6 @@
 module Api
   class UsersController < ApiController
     before_filter :load_user, :only => [:show, :update, :destroy]
-    before_filter :require_not_current_user, :only => [:destroy]
-    before_filter :authorize, :only => [:create, :destroy, :ldap, :update]
```
kevinmtrowbridge: I want to pull out everything that's `authorization` related from the `api` component.  We can do this by using decorators.  So, I want to preserve the existing behavior into the `authorization` component.  This means, in the `authorization_simpler` component, I can modify without fear of losing the existing behavior.

components/authorization_simpler/lib/authorization.rb
```diff
@@ -0,0 +1,285 @@
+require "authorization/engine"
+
+module Authorization
+
+  ROLE_PERMISSIONS = {
+
```
kevinmtrowbridge: I think this hash contains most of the information that was contained in the database in the previous incarnation: ChorusClasses, Roles, Permissions, and Operations ...

We lose the ability for admins to edit (on the fly) what operations the roles can do to what objects, but we had no plans to implement this for now, anyways -- and this make it SO MUCH SIMPLER ... we can rebuild this in the future once we've worked this initial part out.

components/authorization_simpler/spec/controllers/api/users_controller_spec.rb
```diff
@@ -0,0 +1,362 @@
+require 'spec_helper'
+
+describe Api::UsersController do
```
kevinmtrowbridge: I pulled these tests out of the Api::UsersController specs -- I think we should try to move the `authorization` related bits out of the `api` specs and into this component.

There are 3 specs failing right now, I know about this but it's not so important right now.

db/seeds.rb
```diff
@@ -28,15 +25,5 @@ def method_missing(*args)
...
-admin_role = Role.find_or_create_by(:name => 'admin'.camelize)
-
-site_admin_role.users << chorusadmin if chorusadmin && !site_admin_role.users.include?(chorusadmin)
-admin_role.users << chorusadmin if chorusadmin && !admin_role.users.include?(chorusadmin)
-
-
-# Add chorusadmin to default group
-default_group = Group.find_or_create_by(:name => 'default_group')
-chorusadmin.groups << default_group
-
```
kevinmtrowbridge: There are a few details here that aren't fully abstracted into the components ... for example if we merged this PR to master, we couldn't go back to the other `authorization` component without undoing these changes to the `$CHORUS_HOME/db/seeds.rb` -- something to remember / figure out.

components/authorization_simpler/db/migrate/20151201216000_create_tables_for_authorization_simpler.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTablesForAuthorizationSimpler < ActiveRecord::Migration
+  def up
+    add_column :users, :roles, :string, array:true, default: []
```
kevinmtrowbridge: @AndrewLngdn: I came up with this simple way for a user to have multiple roles -- http://blog.plataformatec.com.br/2014/07/rails-4-and-postgresql-arrays/

Basically Rails / PosgreSQL provides a special attribute type which is a serialized array.

components/authorization_simpler/app/decorators/models/user_decorator.rb
```diff
@@ -0,0 +1,28 @@
+User.class_eval do
+  # KT: The 'User#roles' attribute is a special Rails Postgres 'array' attribute type ...
+  # http://blog.arkency.com/2014/10/how-to-start-using-arrays-in-rails-with-postgresql/
+
+  # KT TODO: what to do about the scope :admin in core -> user.rb
+  scope :admins, -> { where("'admin' = ANY (roles) OR admin = true") }
```
kevinmtrowbridge: We can quickly run queries on the special serialized array attribute by using this special syntax: `where("'admin' = ANY (roles)")`

http://blog.plataformatec.com.br/2014/07/rails-4-and-postgresql-arrays/

components/authorization_simpler/app/decorators/models/user_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  # NB: Don't modify the 'roles' array directly, use these methods ...
+  # http://stackoverflow.com/questions/17283854/new-data-not-persisting-to-rails-array-column-on-postgres
+  def add_role(role)
```
kevinmtrowbridge: Worth noting -- if you try to do this:

`user.roles << 'admin'` OR `user.roles.push 'admin'`

... after saving, when you reload the record, the changes will not have taken.  It fails silently.  Details in the post I linked to.

components/authorization_simpler/app/decorators/controllers/api/users_controller_decorator.rb
```diff
@@ -0,0 +1,36 @@
+Api::UsersController.class_eval do
+  before_filter :authorize, :only => [:create, :destroy, :ldap, :update]
+
+  private
+
+  def authorize
+    unless passes_controller_authorization? || passes_role_based_authorization?(@user)
```
kevinmtrowbridge: Before, you were dividing the type of authorization into two: 'legacy_authorization' and the new type ...

But I don't think we'll ever be able to get rid of the legacy authorization, because, for example, obviously current_user can update themselves -- how would we track that in the role based authentication?  I see no obvious way.

So, I've renamed them:

'legacy authorization' -> controller_authorization
'new system' -> role_based_authorization

In general, I suggest we create a `passes_controller_authorization?` method for each Api controller via a decorator, and then we can add a singular `passes_role_based_authorization?` method in the ApiController decorator.

components/authorization_simpler/app/decorators/controllers/api_controller_decorator.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def passes_role_based_authorization?(obj)
+    return false unless obj
+
+    all_actions_allowed_for_user = []
+    current_user.roles.each do |role|
+      all_actions_allowed_for_user + Authorization::ROLE_PERMISSIONS[role][obj.class.to_s]
+    end
+    all_actions_allowed_for_user.include? params[:action]
+  end
```
kevinmtrowbridge: Here's a simple way of uniting the permissions for all of a users' multiple roles ...

kevinmtrowbridge:
(Another thing I realized is that, we should probably rename this component to be called 'permission' or something similar because, authorization is too close to authentication.  I mis-typed several times while commenting.  Also, in the near future I want to create an `authentication` component and it would suck to have them named so similarly.)

kevinmtrowbridge:
@AndrewLngdn: OK -- I think I'm done for now.  I will move on to the Chiasm work now.  Let's discuss this tomorrow / after you've had a chance to digest it all?

Thanks,
KT

PR #216, opened 2015-12-10
DEV-13450 remove DS password requirement for hive JDBC connection wit…

AndrewLngdn:
…h kerberos

Changes so that a 'kerberized' data source doesn't show user-pass when editing. Current behavior is to send 'default'/'default' as the password, so I just hid the fields and set the password to 'default' when saving.


alpinebuilder:

PR #218, opened 2015-12-16
Fix API specs

AndrewLngdn:


alpinebuilder:

components/api/app/controllers/api/data_sources/members_controller.rb
```diff
@@ -6,7 +6,9 @@ class MembersController < ApiController
 
     def index
       accounts = DataSource.find(params[:data_source_id]).accounts.includes(:owner).order(:id)
-      accounts = DataSourceAccount.filter_by_scope(current_user, accounts) if current_user_in_scope?
+
+      # TODO: remove scope filtering from data source accounts
+      DataSourceAccount.filter_by_scope(current_user, accounts) if current_user_in_scope?
```
kevinmtrowbridge: @AndrewLngdn: did you mean to remove `accounts = ` from this line?

kevinmtrowbridge: This is the only, actual "app code change" in this PR, it's fairly safe to merge, just curious about that question, before merging.

kevinmtrowbridge:
Woot!  I pushed a few more fixes as well.  We are down to like 4 failing specs in API!!!!!

Andrew just one question for you, then let's merge this in!

kevinmtrowbridge:
The API specs are all fixed now, so this is no longer needed.  Thanks for your work, @AndrewLngdn.  Closing.

PR #229, opened 2016-01-08
RGP 5.9: authorization_simpler sub-feature branch

AndrewLngdn:
With tests!


alpinebuilder:

components/authorization_simpler/app/mixins/authorization/permissioner.rb
```diff
@@ -0,0 +1,13 @@
...
+# that generate bitmasks or permission bits using the PERMISSIONS
+# constant on the class
+
+module Authorization::Permissioner
+
+  # Returns true if user has site wide admin role.
+  def self.is_admin?(user)
+    admin_roles = %w(site_administrator application_administrator admin application_manager)
+    (user.roles & admin_roles).any?
+  end
```
kevinmtrowbridge: I feel like we will have problems completely removing the concept of "admin" from core.  I believe I tried to do that when I created these components and I failed.  It also seems like one of those things that you need in the core (like a "root" account).

If that's the case then we might solve the problem by renaming the "admin" that's in the core, and keeping that as a boolean.  We could rename it to "superuser" "root" "core_admin" ...

Then we could handle these other "admin" concepts (that have to do with RGP) in an easier way.

When I read this method that's what I start thinking about, how it's confusing to have "admin" done in two different ways with the same name: with a boolean on User, and via the User's roles.  We should figure that out before going too far down this road.

components/authorization_simpler/lib/authorization.rb
```diff
@@ -282,4 +282,15 @@ module Authorization
...
+  CLASS_MAP = {
+      'GpdbDataSource' => 'DataSource',
+      'GnipDataSource' => 'DataSource',
+      'HdfsDataSource' => 'DataSource',
+      'OracleDataSource' => 'DataSource',
+      'PgDataSource' => 'DataSource',
+      'JdbcDataSource' => 'DataSource',
+      'JdbcHiveDataSource' => 'DataSource',
+      'HdfsDataSource' => 'DataSource'
+  }
```
kevinmtrowbridge: Maybe there is a magical way to handle this with Ruby inheritance.  Since these all inherit from DataSource, we could just define a method in data_source.rb:

```
def authorization_clazz
  DataSource
end
```

We can do this in a `authorization_simpler/app/decorators/models/data_source_decorator.rb` file ...

components/authorization_simpler/app/decorators/controllers/api_controller_decorator.rb
```diff
@@ -11,18 +11,49 @@ def initialize(message, action, subject, payload=nil)
...
+      # Helps map subclasses to the permissions class (ex: GpdbDataSource -> DataSource)
+      object_class = Authorization::CLASS_MAP[object_class]
+
+      if (object_class.nil?)
+        raise "Error: object class does not have permissions assigned. Check input and make sure the mapping is correct in Authorization::CLASS_MAP\n"
+      end
     end
-    all_actions_allowed_for_user.include? params[:action]
+
+    Authorization::ROLE_PERMISSIONS[role][object_class]
```
kevinmtrowbridge: This is where we'd use that `authorization_class` (or clazz, if it's a string):

```
object_class = obj.authorization_class
```

components/authorization_simpler/app/decorators/controllers/api/data_sources_controller_decorator.rb
```diff
@@ -0,0 +1,23 @@
...
+  def passes_controller_authorization?(action, data_source, user)
+    case action
+      when :show
+        return data_source_is_shared(data_source)
+      when :update, :destroy
+        return user_is_data_source_owner(data_source, user)
+      else
+        true
+    end
+  end
```
kevinmtrowbridge: I love it!  You absolutely got what I was going for.  Thank you.

components/api/app/controllers/api_controller.rb
```diff
@@ -46,6 +46,11 @@ def head(status, options = {})
 
   private
 
+  # This method is called if the method is not defined in the authorization_simpler component
+  def authorize!(action, object, user, options)
+    Authorization::Authority.authorize! action, object, user, options
+  end
```
kevinmtrowbridge: Hmm this is just a 'medium-level' comment, it's kind of an implementation detail, but a few thoughts about this:

1) The old Authorization::Authority.authorize! method is never going to be present in the app while the new authorization_simpler component is there.  So this would always throw an exception in any case.

2) I see how in (directly above) data_sources_controller, by doing this you are able to leave the methods inline, in the actions.  

But, if possible I think I prefer the approach I took in users_controller -- using before_filters to handle it and not having any authorization stuff in the controller at all.  

Here you are passing the method along as a symbol (`:destroy`) ... but you can get this directly from `params[:action]`.  

You can handle the options `current_user_is_object_owner` by naming the before_filters and the methods in the specific controller decorator in a creative, easily legible way.  

This is better because we can just comment out the Gem in the Gemfile, and boom, we're done, all traces of it are gone.  We can see all parts of it by working within the authorization_simpler component ... it will be less magical and more legible.

components/api/app/controllers/api/data_sources_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
     end
 
     def destroy
-      Authorization::Authority.authorize! :destroy, @data_source, current_user, {:or => :current_user_is_object_owner}
+      authorize! :destroy, @data_source, current_user, {:or => :current_user_is_object_owner}
```
kevinmtrowbridge: If we run into load-order problems with the `@data_source` not being defined when using before_filters, we can use `prepend_before_filter` or `append_before_filter` to adjust the load order to be correct.

components/api/app/controllers/api/data_sources_controller.rb
```diff
@@ -41,7 +41,7 @@ def update
...
       @data_source.destroy
       head :ok
     end
```
kevinmtrowbridge: More comments about this file, we should also move the `hide_disabled_source`, `hide_data_source?` and the before filter into the authorization_simpler component.

components/authorization_simpler/lib/authorization.rb
AndrewLngdn: I like the authorization_clazz method better than checking inheritance. In the old system, there was a fancy method that walked up the .ancestors tree until it found the right authorization class. You can't just check the first ancestor because many objects have ancestors in between. Authorization_clazz would fix that

kevinmtrowbridge: "walking ancestors tree" ... I think this might be too magical.  The explicit `authorization_clazz` is basically just terser, I like that it has less code.

components/api/app/controllers/api_controller.rb
AndrewLngdn: I don't think we'll be able to get away with just using the action name (at least, not with the original RGP specs Milano and Sara made a while ago). For example, look at the workspace permissions in authorization.rb. edit_settings and add_members would both fall under the 'update' actions, but the system needs to be able to differentiate between the two. Not sure if we're keeping those permissions or not though.


The reason I didn't use the 'before' filter was to maintain the ''swapability" of this component. It looks like most authorize calls could be easily switched to filters because most of the pre-authentication work is simply a lookup except in certain cases. Changing these to filters would involve updating every controller to use instance variables, and certain methods would have to be split in half. These splits would look arbitrary from the perspective of the API component. It might be more trouble than it's worth given that we can just override the authorize call with a decorator and redirect the flow through our new logic

components/authorization_simpler/lib/authorization.rb
AndrewLngdn: should this be a model decorator in the authorization_simpler component?

kevinmtrowbridge: Yes exactly, we would put it in `authorization_simpler/app/decorators/models/data_source_decorator.rb`

components/api/app/controllers/api_controller.rb
AndrewLngdn: I'm going to go the route of not changing the method signature. A lot of the newer and more interesting permissions won't map to controller methods, so we'll need to figure out what action is being performed and pass that into the authorize call (ex: [edit_settings add_members delete_members add_to_scope] for a workspace would all fall under "update")

components/api/app/controllers/api/data_sources_controller.rb
AndrewLngdn: Hmm but is hiding disabled data sources authorization? I'm going back and forth on this one

components/api/app/controllers/api_controller.rb
kevinmtrowbridge: I understand what you are saying.  I agree that the filters could get to be weird.  We can also experiment with using the splat operator `*args` to allow a variable number of parameters to be passed.

In the root ApiController we can put:

```
def authorize(*args) # allow any (including 0) number of arguments
  # Override me in the Authorization component
  true
end
```

Then, the Authorization component, we can decorate the ApiController:

```
def authorize!(*args)
  user = args[0]  
  object = args[1]  
  action = args[3] || params[:action] # pass in nil to get to options if necessary
  # etc
end
```

In the individual controllers we can further override IF necessary:

```
def authorize!(*args)
  # super special override
end
```

... something like this.

components/api/app/controllers/api/data_sources_controller.rb
kevinmtrowbridge: Yes it is, because only certain types of users can do it.  This fine-grained level of authorization control is exactly what this new format (of using decorators) allows.  We should liberally move everything out of API that has to do with Authorization.

One thing we have to keep track of though, is that, if we want to maintain the backwards compatibility with the old Authorization component, we have to move this stuff in there too.  Just create the same controller_decorator in there, but with only these methods.

AndrewLngdn: Not sure I agree. Does that mean we should move the destroy method to the authorization module, because only certain users can destroy a data source? Also, the method 'hide_disabled_source' is a filter, not an action. Disabling is done via the update controller

kevinmtrowbridge: But @AndrewLngdn: at the core of these methods is this call:

```
    def hide_data_source?
      !Authorization::Permissioner.is_admin?(current_user) && @data_source.disabled? && @data_source.owner != current_user
    end
```

... where we are hiding the disabled data source based on whether or not someone is an admin?  When we are done creating the `authorization_simpler` component, we should be able to delete it, completely remove it from API, and API should still be able to function.  So, we can't have the api/data_sources_controller talking about admin at all, because it will throw an exception.  (Unless we come up with some way of having a core `superuser` boolean that's independent from admin.)

I think disabled data sources should be shown by default to everyone when an authorization component is not present.  The authorization component adds the sophistication that some are hidden from non-owners and non-admins.  Only owners and admins are authorized to see the disabled data sources.

AndrewLngdn: I see what you're saying with regards to the is_admin? part of the functionality. But the other part of the boolean is core to the minimal behavior of a disabled data source: `@data_source.disabled? && @data_source.owner != current_user`. 

Note that this isn't filtering the disabled data sources from the index (they are not filtered by default) , this is hiding the contents of a data source when it's disabled, so that an arbitrary user can't browse a disabled data source. This functionality should be present no matter what authorization system is in place, if any. Let's try to figure out how this should change later

components/api/app/controllers/api_controller.rb
AndrewLngdn: This seems like a decent path to start down. I'll start to make the changes and we'll see how we like them

components/api/app/controllers/api/data_sources_controller.rb
kevinmtrowbridge: I see what you're saying. Maybe split it in half? It's ok, we can review all this once more when we've dealt with more situations, we'll probably get better / more opinionated. Thanks for engaging in the discussion with me.

components/authorization_simpler/app/decorators/controllers/api/dataset_imports_controller_decorator.rb
```diff
@@ -0,0 +1,20 @@
+Api::DatasetImportsController.class_eval do
+  include WorkspaceAuthorization
+
+  private
+
+  def passes_controller_authorization?(action, object, user)
```
AndrewLngdn: @kevinmtrowbridge So I've been porting over a lot of these controllers and I'm beginning to think that our pattern isn't quite perfect (although we are going the right direction!). 

I think 'passes_controller_authorization?' is sortof a misnomer. We're really authorizing with two ideas: object-level permissions based on current user, and class-level permissions based on current_user's role. These two checks almost mirror our 'authorize!' method in the ApiController decorator:

    passes_controller_authorization?(action, object, user) || passes_role_based_authorization?(action, object, user)

'passes_controller_authorization?' isn't quite the right name because the object-state we're checking is not really tied to the controller at all. 

Look at the :show case below. We're in the DatasetImportsController, but the permission we're checking has nothing to do with DatasetImports. In fact, the :show action comes from this authorize! call: 

    authorize! :show, workspace, current_user

And it happens in the #index action of the ImportsController. Which makes sense: you can't see a workspace's dataset imports unless you have :show permissions for the workspace. 

By tying these methods to the controller, we're likely to smear object-level checks over the controllers, when they should really be grouped by the object in question.  

Just something to think about. I'm going to continue with the same name and scheme that I've been following. Renaming the method is a simple find-and-replace, and re-organizing the object-level checks would be fairly simple once all the controllers are done, I could easly do it in a clean-up period down the line. It would actually take us closer to how allowy was organized, plus roles :)

PR #230, opened 2016-01-12
Add descriptions to jobs and workfile

AndrewLngdn:


components/frontend/app/assets/javascripts/templates/job_show_content_header.hbs
```diff
@@ -9,6 +9,12 @@
 <div class="job_information">
 <div class="item_metadata">
     <div>
+        <div class="description">
+            {{#if description}}
```
jleealpine: Move {{#if description}} to between line 11 and 12 so that it also covers the <div class="description"> line, like in job_item.hbs.  (or vice versa, the point is that these two probably should be consistent, unless there's a particular reason why they're different)

PR #233, opened 2016-01-15
Delete the old Ruby installer

kevinmtrowbridge:
In 5.5, @haozou rewrote the ruby installer in Python.  I believe it was a mostly straightforward port, the new installer has been well accepted and, it's certainly faster, and it has also acquired some additional niceties (some of which I have worked on).  So, I think it's time to delete the old installer, because it has very similar names to the new installer, and it's confusing when working with the new installer to have a
bunch of old files that are never executed, lying around.

#cleanup

-----

@haozou can you review this?  I have to do a bit more testing before I merge it in, but hopefully I can do this early next week.  Thank you.


haozou:
looks good to me

kevinmtrowbridge:
I think it's time for this one.  Merging it in.

PR #235, opened 2016-01-22
Remove "honor_codes" gem

kevinmtrowbridge:
honor_codes is a ruby gem by ex-Chorus developer Robbie Gill, that was used for the old (base64 encoded) way we were doing the licensing.  As part of the work @jleealpine and I did for 5.8, we moved to a different way, that uses a new gem (which I created) called "SignedConfig" ... so here, just cleaning up and removing honor_codes from the project, because it is not used any more.

#cleanup


kevinmtrowbridge:
Ok -- these "cleanup" branches are ready to go.  I will merge in some that are obvious.

PR #236, opened 2016-01-22
Cleanup licensing

kevinmtrowbridge:
In the old way of doing licensing, one thing you could "pay extra" for, was "full search" -- but, we're not doing licensing this way any more.

@jleealpine: let's discuss with Lawrence and with product to hear if there are any plans to ever use these ways of limiting functionality again:
home_page, limit_sandboxes?, limit_data_source_types?, limit_workspace_membership?

... otherwise we should delete them.  Unless we have plans to bring this back, we should delete the code that made that work.  Otherwise, it's confusing and it will decay in time.  In the future, no one will ever be more aware that this code is no longer ever being executed, than we are now.


jleealpine:
In general, looks good, since it's a removal check-in.  But yes, let's continue the conversation about what will be good to remove going forward as well.

kevinmtrowbridge:
OMG Travis worked and found 3 failing tests in Core and 2 in Frontend.  I will fix these and try again.

PR #243, opened 2016-02-03
authorization_simpler: KT enabling tests and moving controllers over

kevinmtrowbridge:
Controllers in subdirectories starting from z -> a, Andrew is doing a -> z.  We will meet in the middle ...


spec/factories/users.rb
```diff
@@ -17,7 +17,5 @@
     sequence(:first_name) { |n| "Admin_#{n}" }
     sequence(:last_name) { |n| "User_#{n}" }
     admin true
-
-    after(:build) { |user| user.add_role('admin') }
```
AndrewLngdn: We actually need this line,  the 'admin' role is how admin authentication works in the new system

components/api/app/controllers/api/workspaces_controller.rb
```diff
@@ -94,11 +94,6 @@ def update
       workspace.attributes = attributes
       Authorization::Authority.authorize! :update, workspace, current_user, {:or => :current_user_can_update_workspace}
 
-      if workspace.changed.include?("owner_id")
```
AndrewLngdn: So happy to be rid of this stuff. When this branch is merged in I'll do another pass over the codebase to get rid of more Role stuff

components/api/app/controllers/api/licenses_controller.rb
```diff
@@ -11,6 +11,7 @@ def reload
       present lic
     end
 
+    # KT TODO: is this, to override the "require_login" before_filter? or is it cruft?
     def require_login
     end
```
kevinmtrowbridge: @jleealpine: I will come by to chat about this with you.  I wonder if I can delete this method?

spec/factories/users.rb
kevinmtrowbridge: Right, thank you for remembering this.  I will put this back and then, merge this in.

PR #244, opened 2016-02-03
AuthorizationSimpler: remove old scoping

kevinmtrowbridge:
The old Authorization component had the beginnings of a scoping system, but it
was never actually wired up.  It was always disabled at its root.  It's confusing and
spammy -- here I excise all traces of it.
    
#rgp_5.8 #authorization_simpler #cleanup

@AndrewLngdn please take a look ...


AndrewLngdn:
Looks good

PR #245, opened 2016-02-04
Re-gitignore fixtures: sorry

kevinmtrowbridge:
Hi Chorus team, I am sorry that I added the fixtures to git, that was a mistake.  I have a better handle on the fixtures now after all the work I did to Dockerize / Travis-ize the tests.  I would like to remove them now.  Does anyone object?

It really is a pain the way they always turn up in `git status`


kevinmtrowbridge:
Ok, I don't think there will be any really opposition to this.  Sorry for that failed experiment -- I will merge this into master now.

PR #246, opened 2016-02-04
AuthorizationSimpler: cleanup roles hash

kevinmtrowbridge:
Hi @AndrewLngdn this is a small one but, just cleaning a few things up that I noticed:

"In the new Authorization component, these models (ChorusScope etc) no longer exist ... scrubbing them out. #cleanup"


AndrewLngdn:
Looks good to me. We can re-add anything that we need to very easily

kevinmtrowbridge:
Great, I agree, merging.

PR #247, opened 2016-02-04
AuthorizationSimpler: cleanup api, Api::AccessDenied, & workspaces/csv_imports_controller

kevinmtrowbridge:
* We can't use any "Authorization" namespaced stuff within the Api component.  In order to call into our
new `authorize!` method in the ApiController, we need to just simply call `authorize!` not, `Authorization::Authority.authorize!` (that is left over from the old way).
* Similarly, we cannot use the `Authorization::AccessDenied` exception inside of the Api component, but we need something like this, so I create a new (identical) exception called `Api::AccessDenied` ...
* Finally with these fundamental things rectified, moving the Workspaces/CsvImportsController over to the new system ...

@AndrewLngdn: please take a look.


AndrewLngdn:
Looks good to me

kevinmtrowbridge:
Thanks!

PR #248, opened 2016-02-05
AuthorizationSimpler: api independence

kevinmtrowbridge:
Andrew: here's the bulk of the work (in draft form) to get the Api code (not tests) to not depend on AuthorizationSimpler.  If you run the Api tests in this branch, ~450 will fail, I believe.  

If we can delete / migrate the Api tests that are failing (because there's no authorization happening in the Api component anymore) into the AuthorizationSimpler component, this will a be a pretty good endpoint for our refactoring.


components/core/app/services/workspace_members_manager.rb
```diff
@@ -34,7 +34,7 @@ def update_membership
   def remove_members_with_roles(removed_members)
```
kevinmtrowbridge: @AndrewLngdn: WorkspaceMembersManager -- I believe it belongs in Core.  I moved it but, I have not edited it to remove the parts that deal with the old roles system.  So essentially, those parts just need to be deleted.  Its current problems are going to turn up in failing tests ...

kevinmtrowbridge:
Hi @AndrewLngdn: ok this is what I have ... really this is more of a sketch ... many tests will be broken in Api.  Please ask me if there are fundamental "componentization" parts that just, seem wrong.

AndrewLngdn:
Looks good to me, I'll merge it in and get to work

PR #250, opened 2016-02-09
Docker tests: access Jasmine server inside the container

kevinmtrowbridge:
I was working with @mmilano on some broken Jasmine specs, and, we wondered how to access the Jasmine server inside of the Docker container.  So I did a little work to figure this out.

@mmilano -- what you do is just:

    docker-compose run --service-ports web ./components/frontend/run_jasmine_server.sh

Then, once it starts, you can access the Jasmine server at a URL like this:

    http://<docker-machine ip>:8888?spec=<my failing Jasmine spec name>

So, one example for me is:

    http://192.168.99.100:8888/?spec=chorus.dialogs.ManageJoinTables

Now, if you want to edit files, for now, the easiest way to do this is to ssh into the docker image and use vi:

So, figure out the name of the running container: 
* `docker ps` ... look for something like `chorus2_web_run_3` ... then
* `docker exec -it chorus2_web_run_3 bash`

Now, inside the container, just edit the file (for the one we were looking at before: `/chorus/components/frontend/spec/javascripts/dialogs/dataset/manage_join_tables_dialog_spec.js`, then reload the specs (using the browser with the filter string) ... and you will see the effect of the change.

For doing 'development' with docker, later we can setup a 'volume' so that, the filesystem inside the docker container is mirrored with one on the developer's OSX machine.


kevinmtrowbridge:
This work has been incorporated into master by other means.  Closing this PR.

PR #255, opened 2016-02-16
Add heatmap, time series, and box plot visualizations for Hadoop csv file preview

imccarter:
dataUpdated and the (un)hideVisualization functions for each visualization could use another pair of eyes. Some bugginess occurred when checking or unchecking the "first row contains header" option and prompting re-renders that I've handled in those functions; standardizing and making sure they're entirely necessary may be a good idea.


midnighteuler:
@imccarter - You'll want the PR to be into master, not the 5.8 branch.

nkapinos:
@imccarter Do you need help to merge this into Master instead of 5.8?

PR #258, opened 2016-02-18
DEV-13998 update workfiles controller to use data source auth

AndrewLngdn:
@midnighteuler This fixes a bug where a workfile couldn't be opened because the auth logic was checking the wrong authorization code. The bug was due to that extra decorator stuff that I was telling you about yesterday. I'm going to clean it out during the cleanup period after this release, so it should be good to merge.


midnighteuler:
Looks fine to me.

PR #260, opened 2016-02-23
DEV-14045, DEV-14043

imccarter:
DEV-14045 More informative error messages for visualizations.
DEV-14043 Fix visualizations getting stuck when switched during initial load.


jmlewis:
LGTM

PR #262, opened 2016-02-26
Dev 13839 worklets in license

mmilano:
@jleealpine could you please do a code review on the changes in this branch.


jleealpine:
I think these changes look good, in general I wasn't able to find too many specs that seem to depend on touchpoint.enabled, so I think we're probably good to merge this in.

jleealpine:
I done messed up, I forgot about some of the iffyness with api_docs that requires some stubbing in some of the spec_helper and fixture_builder files and it ended up breaking the build.  I checked in one more change here that fixes it: https://github.com/Chorus/chorus/commit/ec50a6cbb5eeee138de4420b6996ff39ec931f70

PR #268, opened 2016-03-09
KT MERGING SOON: RGP refactor

AndrewLngdn:
Move permissions check from controller decorators to the new 'permissions' folder. This helps reduce a ton of duplicated code (ex: check user_is_member? of a workspace in the WorkspacePermissions class instead of in each controller decorator that authorizes a workspace object). 

This cut out 50 or so controller decorators, and simplifies our auth code path. Permissions are now closer to the model being inspected and not scattered over the controllers (ex: workspace auth was happening the same way in 20 different controllers). 

@kevinmtrowbridge let me know if you think any changes should be made


kevinmtrowbridge:
@AndrewLngdn: I just merged master, it was easier and you won't need to fiddle around.  I will review it once the tests run.

kevinmtrowbridge:
Hi @AndrewLngdn: I am reviewing this now.  Thanks for your patience.

components/api/app/controllers/api/data_sources/accounts_controller.rb
```diff
@@ -28,10 +28,10 @@ def updated_account
 
       account = data_source.account_for_user(current_user) || data_source.accounts.build(:owner => current_user)
 
-      authorize! :edit_credentials, account, current_user, { :or => :current_user_is_account_owner }
+      authorize! :edit_credentials, account, current_user
       account.attributes = params[:account]
 
-      authorize! :edit_credentials, account, current_user, { :or => :current_user_is_account_owner }
+      authorize! :edit_credentials, account, current_user
```
kevinmtrowbridge: Hmm I think the duplication is essentially a sort of typo, here.

components/api/app/controllers/api/attachments_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
 
     def show
       attachment = Attachment.find(params[:id])
-      authorize! :show, attachment.note.note_target, current_user, {:or => :handle_legacy_show}
+      authorize! :show, attachment.note.note_target, current_user
```
kevinmtrowbridge: In cases like this, it might make sense to expand it a bit:

```
event = attachment.note.try(:note_target)
authorize! :show, event, current_user
```

... and I generally try to be slightly proactive to guard against nils.  Chorus is general is pretty dependent on the data being, "just so," otherwise it will start throwing exceptions right & left.

components/api/app/controllers/api/data_sources/members_controller.rb
```diff
@@ -60,7 +60,7 @@ def update
 
     def destroy
       gpdb_data_source = DataSource.find(params[:data_source_id])
-      authorize! :update, gpdb_data_source, current_user, { :or => :current_user_is_object_owner }
+      authorize! :update, gpdb_data_source, current_user
```
kevinmtrowbridge: This is just a note, because I'm pretty sure the code is correct here, but in places where you only need `gpdb_data_source` to be a local variable, it might be easier to read if you name it more distinctly.  

`@gpdb_data_source` and `gpdb_data_source` are quite similarly named but they're totally different, if that makes sense.

For example, if it's just a local variable you could just call it `data_source`.

components/api/app/controllers/api/data_sources/members_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
 
     def create
       @gpdb_data_source = DataSource.unshared.find(params[:data_source_id])
-      authorize! :update, @gpdb_data_source, current_user, { :or => :current_user_is_object_owner }
+      authorize! :update, @gpdb_data_source, current_user
```
kevinmtrowbridge: For that matter one of the great mysteries of this controller is why `gpdb_data_source` -- are we certain that `@gpdb_data_source = DataSource.unshared.find(params[:data_source_id])` ? How so?  This is probably outside of the scope of `authorization` though.

components/api/app/controllers/api/gnip_data_sources_controller.rb
```diff
@@ -31,7 +31,7 @@ def update
 
     def destroy
       data_source = GnipDataSource.find(params[:id])
-      authorize! :update, GnipDataSource.find(params[:id]), current_user, {:or => :current_user_is_object_owner}
+      authorize! :update, GnipDataSource.find(params[:id]), current_user
```
kevinmtrowbridge: Why not use the data_source we just found?

components/api/app/controllers/api/data_sources/accounts_controller.rb
AndrewLngdn: I was thinking the same thing, it's unlikely that assigning the params would cause the second auth to fail

components/api/app/controllers/api/gnip_data_sources_controller.rb
AndrewLngdn: Should be the same, right? I'm pretty sure we're tossing Gnip support though.

kevinmtrowbridge:
Ok @AndrewLngdn -- I created another PR for the very spammy 'delete old authorization' component work: https://github.com/Chorus/chorus/pull/269, please review that one first and merge it into this one, once its Travis tests pass.

Besides for the comments I made above, I think this is ready to be merged in.  If you want to respond to any of those comments, please do.  So:

1) Review and merge #269 
2) Review my further changes in this branch and respond to comments
3) Merge this stuff into master!  :)

Thanks for your work I feel quite good about the state of the Authorization component now and, even the idea of componentization: 
* it's so cool that there's only ~340 rspec tests in this component and its tests run in about 2 minutes
* we can upgrade it to Rspec 3 at any time and use any testing libraries we want

KT

AndrewLngdn:
Travis errors seem to indicate a problem with the DataSourceAuth mixin. I vaguely remember that being an area where the components weren't quite separated. We might need a 'dummy' DataSoureAuth in API for the purpose of running the tests in isolation

kevinmtrowbridge:
Yes, it has to do with Api not being quite independent from Authorization.  We will have to re-examine.  I'm not inclined to make things difficult right now.  This is not our focus anymore.  We've gotten it to a place where we can park it until we want to add in some "role based" authorization.

We can look into it tomorrow.  I'm just happy that Travis is working out!

PR #269, opened 2016-03-15
Delete old Authorization component (Merge first, review last)

kevinmtrowbridge:
* Delete old authorization component.
* Ran auto-formatter on new authorization component.


components/authorization/app/validators/user_type_validator.rb
```diff
@@ -1,46 +0,0 @@
-class UserTypeValidator < ActiveModel::Validator
```
AndrewLngdn: This class is only in 'authorization' and isn't in 'authorization_simpler'. When we made the switch to authoriztion_simpler, why did this functionality continue to work? did it?

kevinmtrowbridge: Good question, it may not be working!

components/authorization/spec/controllers/api/members_controller_spec.rb
```diff
@@ -14,7 +14,7 @@
     it_behaves_like "an endpoint that allows admin role" do
       let!(:method) { :get }
       let!(:action) { :index }
-      let!(:params) { { :workspace_id => private_workspace.id } }
```
AndrewLngdn: Looks like lots of the diff came from the autoformatter. I think it was an unwritten rule in chorus to leave whitespace before the first hash element and after the last.

components/authorization/app/validators/user_type_validator.rb
AndrewLngdn: Ok I will look into that tomorrow morning and take care of it

PR #271, opened 2016-03-17
Enhanced tags for 6.0: Dev-14280 / DEV-14285

imccarter:
DEV-14280: add/edit tags option on workspace show page sidebar
DEV-14285: Consistent tag display across chorus; tags added to workspace list items.


components/frontend/app/assets/javascripts/templates/debug_helper.hbs
```diff
@@ -1,11 +1,15 @@
...
- 
+
   if (optionalValue) {
     console.log("Value");
     console.log("====================");
     console.log(optionalValue);
   }
-});
\ No newline at end of file
+});
```
kevinmtrowbridge: I investigated this and found this article: http://stackoverflow.com/questions/17499742/how-do-i-add-console-log-javascript-logic-inside-of-a-handlebars-template

I put `{{debug}}` inside an hbs template and confirmed that it still works.

components/frontend/app/assets/javascripts/dialogs/edit_tags_dialog.js
```diff
@@ -30,12 +30,18 @@ chorus.dialogs.EditTags = chorus.dialogs.Base.extend({
     },
 
     tags: function() {
+        // icm: when editing tags from workspace show page, the workspace model is passed
+        // instad of a collection; this works around that...
+        // TODO: maybe a way to send as a collection to begin with?
+        if(!this.collection && this.model) {
+            this.collection = new chorus.collections.Base([this.model]);
+        }
+
```
kevinmtrowbridge: I found a bug in this area:

### From workspace show page, tags do not refresh, when no tags are present when the page is first loaded

1) Navigate to a workspace show page (for example: http://localhost:3000/#/workspaces/1).  Make sure there are no tags there.  (You can edit and delete them all to get into this state.  If you do this, then reload the page after deleting all tags to recreate the state of there being no tags when the page is initialized.)
2) Click "Add/Edit Tags" -- add two tags.  Notice that the tags do not appear in the WorkspaceSummaryContentHeader.
3) Reload the page, notice that the two tags now show up.
4) Click "Add/Edit Tags" -- add & delete more tags, notice that now they refresh as they should.

alpine chorus-3

I think this has to do with the way that the listeners are wired up in chorus.Mixins.Taggable lines 10-16.  I spent a few minutes trying to fix it up but I haven't been able to figure it out yet so, throwing it back to you, @imccarter.

components/frontend/app/assets/javascripts/templates/item_tags.hbs
```diff
@@ -1,8 +1,7 @@
...
-    {{t "tag_list.title"}}:
-    {{#each tags}}
-        <a class="tag_name" href="{{tagLink}}">{{displayTagMatch .. tagName}}</a>
-    {{/each}}
-</div>
-{{/if}}
\ No newline at end of file
+
+{{!-- icm: awkward double conditional here no expresisons allowed in handlebars --}}
+    {{#if tags.length}}
```
kevinmtrowbridge: Is the `if tags ... if tags.length` necessary?  I created a workspace named "No Tags," then I searched for it and clicked "Show all search results for 'No Tags'" ... I confirmed that this was using this `item_tags` partial to generate the tags by typing `KMT` in the template and confirming that it appeared ... then, I tried testing whether or not there's a case where `tags == true` && `tags.length == false` but, I couldn't easily find this.

You are probably more familiar with this than I am!  But, I didn't easily prove to myself that this is necessary.

components/frontend/app/assets/javascripts/views/dashboard/dashboard_project_list_view.js
```diff
@@ -18,19 +18,33 @@ chorus.views.DashboardProjectList = chorus.views.Base.extend({
...
+
+    // icm: requiredResources only sets the collection's status code; this is
+    // here to set each model's status code for requiredResources.allResponded().
+    // It's somewhat hacky; consider adjusting the base view or fetching.js;
+    // especially if this problem crops up elsewhere
+    validateModels: function () {
+        if (this.collection.statusCode == 200 && this.collection.models) {
+            _.each(this.collection.models, function (workspace) {
+                workspace.statusCode = 200;
+            });
```
kevinmtrowbridge: @imccarter: maybe describe more about how to reproduce the problem that you need to hack around here?  I can take a look or we can ask @midnighteuler or @jleealpine for advice.

components/frontend/app/assets/stylesheets/tags/_tags_simple.scss
```diff
@@ -0,0 +1,26 @@
+/* icm: generalized to use for all places where tags are displayed, but not editable */
+.tags_input {
```
kevinmtrowbridge: Hmm why `tags_input`?  Why not `tags_viewable`?  Is this something leftover from the tag editor?

components/frontend/app/assets/stylesheets/components/_list.scss
```diff
@@ -5,6 +5,7 @@ ul li {
     margin:0;
     padding:0;
     list-style-type: none;
+    @import "tags/_tags_simple"; /* icm: too general? */
```
kevinmtrowbridge: Hmm I see what's really happening here is that, the tags editor `chorus.views.TagsInput` is being initialized whenever tags are displayed ... and so here in certain cases we are hiding that.

I guess I am not super familiar with this but as a first impression, it seems kinda wack to always initialize the editor when we just want to display the tags.  Maybe we could create an option `editable` which would default to false?

So here:
        `this.tagBox = this.tagBox || new chorus.views.TagBox({ model: this.model, editable: true });`

or otherwise it would just display tags and not wrap them inside of an editor input field?  That way we wouldn't have to hide the editor in cases where we jut want to display the tags.

components/frontend/app/assets/stylesheets/tags/_tags_simple.scss
imccarter: @mmilano mentioned that originally the tags were editable from anywhere that they appeared, and that only after did we hide the editing features to streamline it a bit; this is leftover from that I believe. We hide the edit stuff most of the time, but it's not hidden in the add/edit modal. Agreed, it could probably be separated into editable/non-editable tag views.

components/frontend/app/assets/javascripts/views/dashboard/dashboard_project_list_view.js
kevinmtrowbridge: @midnighteuler reached out to me with some thoughts.

Here's our conversation:

> tags uses the "required resources" scheme to establish when it has the things it needs; and it assumes those things are individually fetched and so it inspects the returned HTTP code as though they had ... but the collection of workspaces object is fetched from one endpoint
> and so only the container has a return code
> so he mirrored that code into the children to get the required resources to cooperate

Then I respond:

> I see what you’re saying about the tags being returned as a part of the workspace hash and not via individual http calls and, this causing him to need to make that hack
> seems like it might be better to request the workspace tags separately
> does the API not support that?
> https://10.0.0.204:8443/api/tags/get_the_list_of_tags_for_an_entity.html

Mike Souza
> oh I see; so it seems pretty well developed
> add, delete, index, delete multiple, rename, search
> using that would mean one more ajax query per workspace on that page
> ... but it'd be a lot more sensible in the code

@imccarter have you tried / maybe if you have time, experiment with hooking the workspace tags up to the `taggable` API endpoint, perhaps this would resolve a few of the places where you have had to hack around things?  (Maybe you already tried this, in that case apologies!)

kevinmtrowbridge:
Ok I feel I've reviewed this pretty well now with finding a bug and some good fundamental critique.  

I additionally went through and tried to view every place where you added the `itemTags` partial, I didn't find any problems.

Good work though!  It works great besides for that one bug and those things you might address, thanks for your work, @imccarter!

kevinmtrowbridge:
Ok I took a look at the Travis test results.  The test results in the frontend component look legit to me, we should fix those up before merging this in.  These ones:

https://travis-ci.org/Chorus/chorus/jobs/117547523
https://travis-ci.org/Chorus/chorus/jobs/117547524

(The other ones I am working on fixing myself, I broke those, I will merge in my fix to this branch when I have it ready later today.)

@imccarter: I can help you with running the Jasmine tests if you need help!  Otherwise, take a look at the README file in the root of the Frontend component: https://github.com/Chorus/chorus/blob/master/components/frontend/README.md

docker/build/mods/rails/alpine_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 module Api
   class AlpineController < ApiController
-    before_filter :check_work_flow_enabled, :enforce_alpine_or_chorus_only
+    before_filter :check_work_flow_enabled, :enforce_localhost_or_docker_network_only
```
kevinmtrowbridge: @imccarter: these are the mods to fix the failing tests.  After Travis re-runs, the remaining failures should be unique to this branch.

kevinmtrowbridge:
Hi @imccarter: I merged master into this and I am generally working with it to try and merge it in shortly.

components/frontend/app/assets/javascripts/dialogs/edit_tags_dialog.js
kevinmtrowbridge: I confirmed this is fixed, thanks @imccarter!

PR #274, opened 2016-03-19
Default profile picture for Chorus

AndrewLngdn:
Adds default profile pictures to new users based on first and last initial


AndrewLngdn:
Green build!!

kevin-alpine:
Woot!

components/api/app/controllers/api/user_images_controller.rb
```diff
@@ -1,6 +1,13 @@
 # KT TODO: does this file work, since the ImagesController is broken?
 module Api
   class UserImagesController < ImagesController
+
+    def destroy
+      UserImageGenerator.create_profile_image_for(@entity)
+      authorize! :update, @user, current_user, nil
+      present @entity.image, :content_type => 'text/html'
```
mmilano: is this the content type for the svg file?

AndrewLngdn: I copied the content type of https://github.com/Chorus/chorus/blob/master/components/api/app/controllers/api/images_controller.rb#L9, although it looks like it should actually be 'text/json'. SVG is never sent over the wire, it only resides in the backend.

PR #275, opened 2016-03-23
Add user roles for individual workspaces

AndrewLngdn:
Making PR to see what Travis thinks about this branch. Do not merge


components/core/app/services/workspace_members_manager.rb
```diff
@@ -2,58 +2,50 @@
...
     end
 
-    create_events(new_members)
+    create_events(new_members.map{ |m| m[:user_id] })
     @workspace.solr_reindex_later
   end
 
   private
 
   def remove_members_with_roles(removed_members)
```
jleealpine: Is this method still being used anywhere? Or are there future plans for it to be kept around?

components/core/app/models/membership.rb
```diff
@@ -3,9 +3,19 @@
...
 # without having a membership.
 class Membership < ActiveRecord::Base
+  ANALYTIC_MANAGER = "Analytic Manager"
+  BUISNESS_MANAGER = "Business Manager"
+  DATA_SCIENTIST = "Data Scientist"
+  DATA_ANALYST = "Data Analyst"
+  DATA_ENGINEER = "Data Engineer"
+  APPLICATION_DEVELOPER = "Application Developer"
+  OPS_ENGINEER = "Ops Engineer"
+  PROJECT_MANAGER = "Project Member"
```
jleealpine: Please make the field name consistent with the value here, probably PROJECT_MEMBER, I imagine (I was staring at this for an embarrassingly long time thinking "wait, so everyone defaults to a project manager? unless project manager is correct, which I don't think makes sense, but I guess that's someone else's call).  Also be sure to fix all its usages as well.

components/core/app/models/membership.rb
```diff
@@ -3,9 +3,19 @@
...
+  DEFAULT_ROLE = PROJECT_MANAGER
+
   attr_accessible :role
-  belongs_to :user, :touch => true
-  belongs_to :workspace, :touch => true
+  belongs_to :user, :touch => true, :inverse_of => :memberships
+  belongs_to :workspace, :touch => true, :inverse_of => :memberships
 
   validates_presence_of :user
   validates_presence_of :workspace
```
jleealpine: can we add a validator for role that validates presence (unless role can be blank) and makes sure that its value is one of the role types listed above?

jleealpine:
I think the rest of it looks pretty fine.  I guess it says that this is still under dev, so I won't say you can merge, but other than the comments I made, it looks pretty good.

AndrewLngdn:
Great comments @jleealpine! thanks for the review

components/core/app/services/workspace_members_manager.rb
AndrewLngdn: I think it still creates notifications and events

PR #278, opened 2016-03-29
Email component 6.1

kevinmtrowbridge:
**I also put it on my docker instance, you can access it here: http://10.10.6.167:8080/#admin/email_configurations/1**

Documentation JIRA issue: https://alpine.atlassian.net/browse/DOC-170
Feature JIRA issue: "Welcome Email when a new person is added to the application"
https://alpine.atlassian.net/browse/DEV-14257

-----

* Upgrade rails from 4.1.10 -> 4.1.15 for security patches and a 'preview email' related patch that I need.
* Move all email related code into a new "email" component.
* Parse existing email configuration out of the chorus.properties file, if it exists, and put it into the database.
* Add Admin panel for "Email configuration" -- admin panel has the ability to reload the email configuration on the fly, partially in order to enable "Send test email" functionality.
* Exposing "preview your email templates visually" which was added natively to Rails 4.1 (makes developing emails MUCH easier).
* General cleanup.

The idea here is to expose a way forward for "niche" functionality in Chorus.  When we need to work with LDAP, or Kerberos, etc, we can move that into a component.  Then, we can have an admin panel to make it obvious to ourselves and to customers, whether or not this functionality is present and properly configured.  We can also whittle down the contents of `chorus.properties` to contain only the geekiest things.

Since @mmilano asked us to add a new email to the app in 6.0, I needed to work with the email, so I went ahead and gave it a "tune up," before actually adding the email.  (Make it easy to do the work, then do the work.)


kevinmtrowbridge:
This still needs another two or three hours of work, to give the styling more finesse and make sure everything works properly.  Also we'll need to test upgrades to make sure this won't break email for existing customers.

kevinmtrowbridge:
I created a new PR to merge into this one, that contains the small amount of additional work needed to add the "Welcome email" using this new component.  That PR can serve as a model for adding future emails to the system: https://github.com/Chorus/chorus/pull/279

components/email/spec/dummy/config/environments/test.rb
```diff
@@ -0,0 +1,48 @@
+Rails.application.configure do
+
+  # Custom config options up top:
+
+  # See: https://github.com/Chorus/chorus/commit/0044cca46e8d17ccff6cbcfde3204967a535ddaf
+  ActionMailer::Base.default from: 'Test Mailer <testmailer@example.com>', reply_to: 'Test Replyto <test_replyto@example.com>'
```
kevinmtrowbridge: I can delete this.

components/email/spec/dummy/config/environment.rb
```diff
@@ -0,0 +1,15 @@
+# Load the Rails application.
+require File.expand_path('../application', __FILE__)
+
+# Initialize the Rails application.
+Rails.application.initialize!
+
+Rails.application.configure do
+  config.active_record.mass_assignment_sanitizer = :logger
+
+  config.action_mailer.default_url_options = {
```
kevinmtrowbridge: Should delete this.

components/email/preview_emails.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+./bin/bundle
+./bin/rake app:db:drop app:db:create app:db:migrate
+printf "\n\n\n\n* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * \n"
+printf "Once server has started, visit http://localhost:3000/rails/mailers in your browser to see visual previews of your email templates.\n\n"
```
kevinmtrowbridge: @mmilano: this script makes it easy to use the new `preview_email` functionality in the email component.

mmilano:
@kevinmtrowbridge im a little confused - is this the branch i would pull and revise for a design review? or should i pull the other email branch?

kevinmtrowbridge:
Hi @mmilano: it's this branch -- my apologies for the delay.  It's in a good state now.

IF you pull this to your dev machine, in order for it to work, you must do:

    rake db:migrate
    # then before you switch away, do a
    rake db:rollback # so as to not break your build for other branches

I also put it on my docker instance, you can access it here: http://10.10.6.167:8080/#admin/email_configurations/1

... I'd be curious to configure it to point at other mail servers, I used https://mailcatcher.me/ for development but I have not pointed it at a more serious STMP server.

Also, there's a bug with the docker instance, when you click "Update" you will probably see:

docker_bug

Please just ignore this, it's some technicality I haven't had time to figure out with this deployment method and, I don't think it matters.  **Just reload the page and navigate around this issue for now ...**

email_configuration_panel

kevinmtrowbridge:
To try out the 'welcome_email' feature, configure the email server, and then create a user, with an email address at which you will receive the email.

This is very useful for local testing: https://mailcatcher.me/

We should try it with a real STMP server though.

PR #280, opened 2016-03-29
Jupyter Notebook 6.0 (DO NOT MERGE)

kevinmtrowbridge:
# Using the Python Notebook branch-build at http://10.10.4.95:8080/

## Concepts

* ipython notebooks or ["Jupyter notebooks"](http://jupyter.org/) -- this is basically a web UI for python.  You write regular python in a web UI, and it looks prettier (graphs / images and things).  Also, it's savable, shareable -- it's a common way for data scientists to share work with one another.

* The jupyter executable runs a webserver ... the webserver runs python ... the webserver has an API which Chorus can talk to.

* Chorus is configured to call out to the jupyter API via the `notebook_server.host_url` key in `chorus.properties`.

* Jupyter is configured to call out to the Chorus API (to access databases, etc) via a `CHORUS_ADDRESS` environment variable.

* Jupyter webserver can be managed using a `jupyter` command in bash.

* Jupyter has a concept of `plugins`.  In the ChorusCommander repo, Hao has modified Jupyter by creating a plugin.

## Challenges

It’s really just python, inside a webpage.  

So you can do anything with it: 
* see the entire filesystem, including other people’s files
* start a huge program that hogs all the memory and crashes the computer

So one good solution would be to use this tmpnb “docker” solution (https://github.com/jupyter/tmpnb), which would be a way to run each Jupyter notebook in an “sandboxed” virtual environment.  

However, we are afraid Alpine’s clients won’t be comfortable using docker, because it has some security issues, one of them being that it must be run as root and, apparently it’s quite easy for any process inside of a docker container to escape and be root on the host machine.

This does seem like the best solution though, so one clear way forward would be to, build it out, both ways and give the client a choice:

“Experimental beta feature: jupyter notebooks integration! installed separately” You have two options:
1) accept the security hole of each jupyter notebook running directly on the server
2) install and use docker

Docker security links: 
https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface
http://askubuntu.com/questions/477551/how-can-i-use-docker-without-sudo

## Current state

* Chorus can talk to Jupyter & Jupyter can talk to Chorus.  
* We have the minimum UI built.  
* We have a way to install the Jupyter server to run naively as the chorus user, this does not solve the security issues discussed in the "challenges" section.

The way I see it there are two parts to this work:

1) “easy, but lots of it” — building the UI … 
2) hard, but it can be solved later — figuring out how to install / run / isolate the jyupter server

I think we should push forward with #1 using the branch QA server and give Mike, Hao, Lawrence others more time to figure out how best to proceed 
with #2 (Docker? separate installer?)

I believe @andrewlngdn is also going to help with #1 when he’s feeling better

## Locations

This is the Jira epic for this work:
https://alpine.atlassian.net/browse/DEV-14333

Michael Milano's design documents:
https://docs.google.com/a/alpinenow.com/document/d/1dEj-K97a8mVAPYKcHyUMFzHbfhJYYIrOYDsxxAnAaZ0/edit?usp=sharing

Product / Engineering negotiation spreadsheet:
https://docs.google.com/spreadsheets/d/1KXvxaxkLWVZRxtP9hnD5__IXl94bZQh2cFSm1yzl46U/edit#gid=0

Mike Souza's work to modify Chorus to work with Python notebooks, is in Github, in the Chorus repo, in a branch called 'notebooks': 
https://github.com/Chorus/chorus/pull/280

Hao's work to setup a Jupyter server that can work with Chorus, is in Github, in the alpinedatalabs org, in a repo called 'ChorusCommander':
https://github.com/alpinedatalabs/ChorusCommander

Hao has set up Jenkins to build a version of the Chorus installer that's based off of Mike's branch:
http://10.10.0.215:8080/job/alpine-chorus-packaging-branch/

Hao has setup Jenkins to package ChorusCommander into a separate installer:
http://10.10.0.215:8080/view/Chorus%20notebook/job/chorus_notebook-packaging/

Cuie & Hao have setup an AWS instance, and installed both the chorus-branch-installer and the chorus_notebook installer:
http://10.10.4.95:8080/#notebooks/179

## Managing the instance

In order for the instance to work properly, both Chorus, and the Jupyter notebook must be running.

If you can login to chorus, but you see an error page when creating a notebook, it's probably because the jupyter server is not running.

1. Login, switch to chorus user, and cd to chorus install dir
  - ssh root@10.10.4.95 / alpineRocks
  - su - chorus
  - cd /usr/local/chorus

2. Start chorus
  - ./chorus_control.sh start

3. Start jupyter server
  - export CHORUS_ADDRESS="http://10.10.4.95:8080/"
  - source ./pyenv_paths.sh
  # append the ampersand to run in the background -- the log will still echo to your prompt but, you can do other things, and you can shut down the shell as well
  - jupyter notebook --config=chorus_notebook_config.py & 

4. Stop the jupyter server
  - ps -ef | grep "jupyter" # note PID
  - kill <pid>

## Using the instance

1. Navigate to http://10.10.0.215:8080
2. Login as chorusadmin/admin
3. Navigate to the workfiles tab of any workspace, for example: http://10.10.4.95:8080/#/workspaces/1/workfiles
4. Click "Create New Notebook"
5. Click to "Open Notebook"


components/api/app/presenters/api/notebook_presenter.rb
```diff
@@ -0,0 +1,15 @@
+module Api
+  class NotebookPresenter < WorkfilePresenter
+    def to_hash
+      workfile = super
+
+      workfile[:notebook_server] = {
+          :host_url => ChorusConfig.instance['notebook_server.host_url']
```
kevinmtrowbridge: For future reference, this is where the Jupyter server URL is configured.

midnighteuler: There are example config lines in chorus.properties.example.

Gemfile
```diff
@@ -10,6 +10,8 @@ gem 'vis_legacy', :path => 'components/vis_legacy'
 gem 'vis_chiasm', :path => 'components/vis_chiasm_58'
 
 gem 'hadoopconf_gem', :github => 'Chorus/chorus-hadoop-conf', :ref => 'ce0d34dde0f5f4dd372af6406fafd9c5b2baac5f'
+gem 'chorus_jupyter', :github => 'midnighteuler/chorus-jupyter', :branch => 'master'
```
kevinmtrowbridge: I think -- if it is a legit opensource gem that others might use, by all means you should have it in your personal acct, but if it is a chorus specific gem, maybe we can create a Chorus org repo for it.

Gemfile
```diff
@@ -22,6 +24,7 @@ gem "breakpoint", "~>2.5"
 group :development do
   gem 'license_finder', :require => false
   gem 'pry'
+  #gem 'chorus_jupyter', :path => '/Users/mlsouza/alpine/chorus-jupyter'
```
kevinmtrowbridge: We should remove this.

components/api/app/controllers/api/hdfs/files_by_path_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+        end
+
+        #if params[:id]
+        #  hdfs_entries = HdfsEntry.where(:parent_id => params[:id])
+        #  present hdfs_entries, :presenter_options => {:deep => true}
+        #else
+        #  hdfs_entry = HdfsEntry.find_by_path_and_hdfs_data_source_id('/', hdfs_data_source.id)
+        #  present hdfs_entry, :presenter_options => {:deep => true}
+        #end
+      #rescue ::Hdfs::PermissionDeniedError
```
kevinmtrowbridge: Is this, necessary but incomplete?  Maybe get this to the point where it doesn't have to be commented out before merging.

Gemfile
midnighteuler: I agree; it belongs in the chorus org.

PR #282, opened 2016-03-30
Workspace header icon

mmilano:


components/frontend/app/assets/javascripts/pages/dataset/workspace_dataset_index_page.js
```diff
@@ -58,6 +58,14 @@ chorus.pages.WorkspaceDatasetIndexPage = chorus.pages.Base.include(
         this.onceLoaded(this.workspace, this.workspaceLoaded);
     },
 
+    setupSubHeader: function() {
```
mmilano: @midnighteuler mr souza - im trying to rewrite the subheader for workspaces, and this setupSubheader overrides the one in pages.js.

unfortunately, this override (of which this flagged one is just one)  needs to be added to ALL 9 of the pages that are within the workspace....

what can i do to make this a single method that applies to all of the necessary view pages,
instead of copy and pasting the method into 9 different file?

components/frontend/spec/javascripts/dialogs/workspaces/edit_workspace_dialog_spec.js
```diff
@@ -525,14 +539,21 @@ describe("chorus.dialogs.EditWorkspace", function() {
...
+                            // this.dialog.$("input[name=public]").prop("checked", false);
+                            // this.dialog.$("input[name=public]").prop("checked", false);
+                            this.dialog.$("input[type=radio][name=public][value='public']").prop("checked", false);
+                            this.dialog.$("input[type=radio][name=public][value='private']").prop("checked", true);
+
                             this.dialog.$('form').submit();
                         });
 
                         it("sets the public model attribute to false", function() {
+                            console.log ("test: a");
```
imccarter: Can probably get rid of these console.logs

components/frontend/spec/javascripts/pages/dataset/workspace_dataset_show_page_spec.js
```diff
@@ -1,4 +1,10 @@
 describe("chorus.pages.WorkspaceDatasetShowPage", function() {
+    var workspace;
+    var dataset;
+    var columnSet;
+    var datasetId;
+    var page;
```
imccarter: Do these variables get used?

imccarter:
Left comments, but just for some minor stuff. LGTM.

components/frontend/spec/javascripts/pages/dataset/workspace_dataset_show_page_spec.js
mmilano: @imccarter re: the var definitions. this is something i have been adding to the spec.js files here and there, opportunistically, as i happen to edit those files.
in general, it does no harm, and is correct js practice anyway. 
 i am using it as a sort of experiment to see if it has a beneficial long term effect - the spec suite often will crash and run out of memory because of GC issues in the java heap (i.e. the jruby environment). i was hoping that maybe small effects like this could have some positive impact. probably not, but it doesnt hurt.

components/frontend/spec/javascripts/dialogs/workspaces/edit_workspace_dialog_spec.js
mmilano: done.

PR #283, opened 2016-03-31
Workspace tags 6.0: KT edits

kevinmtrowbridge:
* fixing broken Jasmine tests
* scrubbing Ivan's notes to me
* fixing a Rails 4 upgrade related bug with Workspace caching

Hi @imccarter:

I think all the tests should be passing now.  Take a look through this and, if you agree with everything, first merge this PR into DEV-14280/DEV-14285, then merge the PR for DEV-14280/DEV-14285 into master.

Best,
KT


components/api/app/controllers/api/taggings_controller.rb
```diff
@@ -23,6 +23,7 @@ def create
 
         model.tag_list = tag_names.sort
         model.save!
+        model.touch
```
kevinmtrowbridge: I found another bug while working with the code:
1) Add a tag to a workspace
2) Refresh the page
3) the tag is not there
... do this several times
4) run `rake tmp:cache:clear`
5) refresh the page
6) see all the tags

Sadly the caching code is pretty creaky.  I have made a note to myself to go through and clean it all up when I get the chance.  This solves for that particular case.  The root cause is that the caching uses the `updated_at` timestamps of the models as keys, but in Rails 4 these are only changed when something in the model actually changes.  `touch` will always update the timestamp.

kevinmtrowbridge:
@imccarter -- ignore this test failure, it is fixed in master:

```
1) Api::WorkspacePresenter#to_hash when succinct is true should only present enough for the dashboard list
     Failure/Error: presenter.presentation_hash.keys.sort.should =~ [
```

kevinmtrowbridge:
This missed the train of Ivan's DEV-14280/DEV-14285 branch, so I merged it directly to master.  This PR no longer makes sense, so closing.

PR #287, opened 2016-04-04
Monitoring 6.0: supervisord

kevinmtrowbridge:
This work is to add in the [supervisord](https://github.com/Supervisor/supervisor) "process control system" to Chorus.  This basically goes into the packaging scripts, and adds an extra layer of control between `chorus_control.sh` and the processes.  So, chorus_control.sh, instead of starting and stopping the processes directly, instead tells supervisord to do this.

`supervisord`'s special job is then to watch the processes and keep them running.  It also has a web interface which allows users to view the processes, their logs, restart them, etc.

supervisor status

alpine chorus

## TODO

* Currently I have only gotten this working with a basic "development" environment: it only runs solr, postgres, and mizuno.  We'll need to expand this to work with nginx, jetty, scheduler, workers, indexer, and alpine.  Additionally, the packaging scripts do a lot of extra stuff which we'll need to make sure is still getting done.  We need to get more sophisticated about the integration between supervisord and the packaging scripts.

* Supervisord does not have native functionality to "watch memory" of processes and restart when they use too much, but [there is a popular plugin called superlance](https://github.com/Supervisor/superlance/blob/master/docs/index.rst) which I suggest we explore using.

* Installation of supervisord: on OSX you can just type `brew install supervisor` -- to install on CentOS / production, we'll need to put some python packages into the vendor directory and have the installer run them.  See: http://supervisord.org/installing.html#installing-to-a-system-without-internet-access

* @jleealpine and I will collaborate to push this work further along.  @haozou's input would also be really helpful because, I believe you are most familiar with the packaging system, Hao.

## INFO

* See here a list of the available third party plugins for supervisord, we should explore this: http://supervisord.org/plugins.html


components/admin/app/views/admin/dashboard/index.html.erb
```diff
@@ -44,6 +44,10 @@
         <ul class="list-unstyled">
 
           <li>
+            <%= link_to "Process Control", "http://localhost:9001/", :target => "_blank" %>
```
kevinmtrowbridge: This needs to be a relative path.

packaging/chorus_control.sh
```diff
@@ -65,33 +75,33 @@ function start () {
...
-  fi
-  if should_handle workers;   then
-    $bin/start-workers.sh;
-    EXIT_STATUS=`expr $EXIT_STATUS + $?`;
-  fi
-  if should_handle scheduler; then
-    $bin/start-scheduler.sh;
-    EXIT_STATUS=`expr $EXIT_STATUS + $?`;
-  fi
+#  if should_handle indexer;   then
```
kevinmtrowbridge: I just commented these out for now because, I have not adapted them to work with the new system yet.  As we go along, we can uncomment.

packaging/chorus_control.sh
```diff
@@ -331,9 +341,9 @@ case $command in
         stop ${@}
         start
         ;;
-    monitor )
-       monitor
-       ;;
+#    monitor )
+#       monitor
+#       ;;
```
kevinmtrowbridge: Commented this out, cause this script will be made redundant.

packaging/start-postgres.sh
```diff
@@ -1,31 +1,3 @@
...
-
-if [ -f $POSTGRES_PID_FILE ]; then
-  if kill -0 `head -1 $POSTGRES_PID_FILE` > /dev/null 2>&1; then
-    log "Postgres already running as process `head -1 $POSTGRES_PID_FILE`."
-    exit 0
-  fi
-fi
-
-MAX_CONNECTIONS=$($RUBY $CHORUS_HOME/packaging/update_database_thread_count.rb)
-log "Setting maximum database connections to $MAX_CONNECTIONS"
```
kevinmtrowbridge: We will need to add this back in so that environment variables like `MAX_CONNECTIONS` (which must originate in the chorus.properties file?), get passed along to supervisord.

packaging/start-postgres.sh
```diff
@@ -1,31 +1,3 @@
...
-DYLD_LIBRARY_PATH=$CHORUS_HOME/postgres/lib LD_LIBRARY_PATH=$CHORUS_HOME/postgres/lib $CHORUS_HOME/postgres/bin/pg_ctl -l $POSTGRES_DATA_DIR/server.log -D $POSTGRES_DATA_DIR -w -o "-N$MAX_CONNECTIONS -p$POSTGRES_PORT -h127.0.0.1 --bytea_output=escape" start &>$POSTGRES_DATA_DIR/pg_ctl.log
-POSTGRES_START=$?
-
-if [ $POSTGRES_START -eq 0 ]; then
-    wait_for_start $POSTGRES_PID_FILE
-    log "postgres started as pid `head -1 $POSTGRES_PID_FILE`"
-else
-    log "postgres failed to start, see $POSTGRES_DATA_DIR/server.log for details"
-    exit $POSTGRES_START
-fi
```
kevinmtrowbridge: But other stuff will not be needed at all -- this block, for example, will be done by supervisord and we can delete it.

packaging/start-solr.sh
```diff
@@ -1,34 +1,3 @@
...
-echo "java.util.logging.FileHandler.pattern = $SOLR_LOG_FILE" >> $LOG_CONFIG_FILE
-echo "java.util.logging.FileHandler.formatter = java.util.logging.SimpleFormatter" >> $LOG_CONFIG_FILE
-
-cd $CHORUS_HOME/vendor/solr
-java -Djetty.host=localhost -Dsolr.solr.home=$CHORUS_HOME/solr -Djetty.port=$SOLR_PORT -Dsolr.data.dir=$DATA_DIR -Djava.util.logging.config.file=$LOG_CONFIG_FILE -jar start.jar &> /dev/null &
-solr_pid=$!
-echo $solr_pid > $SOLR_PID_FILE
-log "solr started as pid $solr_pid on port $SOLR_PORT"
-cd $CHORUS_HOME
+supervisorctl start solr
```
kevinmtrowbridge: Under the hood is this command called `supervisorctl` which can be used in a standalone way, for example typing `supervisorctl status` will output something like this:

```
> supervisorctl
mizuno                           RUNNING   pid 30448, uptime 0:21:17
postgres                         RUNNING   pid 30419, uptime 0:21:20
solr                             RUNNING   pid 30430, uptime 0:21:19
```

supervisord.conf
```diff
@@ -0,0 +1,59 @@
...
+;
+; Then, start all services:
+;     supervisorctl start all
+
+
+[unix_http_server]
+file=/usr/local/var/run/supervisor.sock   ; (the path to the socket file)
+
+[inet_http_server]         ; inet (TCP) server disabled by default
+port=127.0.0.1:9001        ; (ip_address:port specifier, *:port for all iface)
```
kevinmtrowbridge: This is the configuration for the web console.  More options are available.

supervisord.conf
```diff
@@ -0,0 +1,59 @@
...
+
+[supervisord]
+logfile=/usr/local/var/log/supervisord.log ; (main log file;default $CWD/supervisord.log)
+logfile_maxbytes=50MB        ; (max main logfile bytes b4 rotation;default 50MB)
+logfile_backups=10           ; (num of main logfile rotation backups;default 10)
+loglevel=blather                ; (log level;default info; others: debug,warn,trace)
+pidfile=%(here)s/tmp/pids/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
```
kevinmtrowbridge: This is the "supervisord dameon" config -- setting things like, the log level, and the location of its own `tmp/pids/supervisord.pid` file ... it also handles logging -- we may be able to use its log rotation & archiving functionality.

PR #290, opened 2016-04-08
DEV-14275 Move workspace list item description to the sidebar

imccarter:
This only affects the workspace list page; when a single workspace is selected, the description will appear on the sidebar, rather than each item containing the description in its body. The description box will be truncated to 10 lines, and will have a "show more" link to expand it.


imccarter:
It'd be trivial to add; but should there be a header/title for the description box? If so, "Summary", or "Description", or something else?

mmilano:
@imccarter heading = "description"

imccarter:
The header is up, and I commented out an errant console.log.

components/frontend/app/assets/translations/messages_en.properties
```diff
@@ -2489,6 +2489,9 @@ workfiles.sidebar.membership.members.one=1 team member
 workfiles.sidebar.membership.members.other={{count}} team members
 workspace.sidebar.membership.owner=Owner
 
+workspace.sidebar.description.header=DESCRIPTION
```
mmilano: labels shouldnt have to be in UC; let the css take care of that with text-transform

components/frontend/app/assets/stylesheets/common/_common.scss
```diff
@@ -128,6 +128,7 @@ $page-margin-bottom:    60px;
 
 // dimensions
 $summary-line-height: 19px;
+$sidebar-summary-height: 193px;
```
mmilano: is the fixed summary height necessary, as opposed to letting it just 'be'?

components/frontend/app/assets/javascripts/models/workspace.js
```diff
@@ -86,7 +86,7 @@ chorus.models.Workspace = chorus.models.Base.extend({
     // map between the internal stagekey and the sequential stage number, via workspaceStages{}.
     stageIndexKey: function() {
         var name = this.get('stage');
-        console.log ("stage index key: " + _.invert(this.workspaceStages)[name] );
+        // console.log ("stage index key: " + _.invert(this.workspaceStages)[name] );
```
mmilano: i propose that you remove this change from this PR, and i (it is my vestige) will remove this in a separate commit.

components/frontend/app/assets/stylesheets/common/_common.scss
imccarter: Was just playing around with it, and the variable above serves the same purpose, but for 2 line truncation. Without it, it just renders to the full expanded length.

imccarter:
Left a comment or made a change according to review notes

imccarter:
@mmilano  I've made it just use the existing variable for line height now; only weirdness is an extra 3px to add to catch the font's descenders, not 100% sure why that was necessary.

mmilano:
@imccarter ha. i was poking at this over the weekend, but didnt finish.
making it variable like you did it probably the best of the options. 

that descender addon is odd - i think it is a consequence of the particular font itself. oh well.

mmilano:
lg.

PR #291, opened 2016-04-13
Adding a set of config shell scripts to add to the supervisord.conf file

jleealpine:
@kevinmtrowbridge: Here are the changes I made.  The good news is that I actually got all of these to start up on a Skytap instance!  It works!!!!! (under very specific circumstances)

Here's what still needs to be done:

- there probably can still be some improvement to the way that the configurations for each service are essentially appended on to the supervisord.conf file.  Still looking up the options we have for this.
- the command line for jetty and alpine are essentially recreations of what is put together by the jetty shell script, which I figured out by reading through the jetty script and experimenting with what it ultimately produces.  While it works, it's a bad idea to depend on this as it might not be a one size fits all solution and it would be a lot cleaner if we could figure out a way to actually get these to work through using the jetty script alone.
- nginx doesn't stop cleanly, it leaves a lot of child processes up and running which causes problems when you try to start nginx up again after that.
- probably some other stuff too, but those are the major known issues

Either way, I'm feeling good that it's at least working in one form or another.  Further work is still needed, but at least we're making progress.


jleealpine:
still plugging away at stuff, fixed the child processes problem, will include in next checkin

jleealpine:
I've committed the files necessary to take care of both the child processes problem, as well as fixing and cleaning up a lot of the messiness/jankiness with all the config files!  Now the supervisord.conf file should be actually able to utilize directly the value of the env variables, significantly reducing the amount of files we need to juggle around (at this point, pretty much all the config-*.sh files can be removed from this branch, but I'll keep them in for now until we get to the point that we need to merge any of this in)

There is still the issue that the jetty and alpine commands need to be revisited to see if there's a less hard-codey way to run those in the foreground.  I'll continue to look into this, I imagine this may take some time...

jleealpine:
One thing I should mention, the start-alpine script actually requires a change on the Alpine side of things as well at the moment.  We can look into a solution that doesn't involve changing up the Alpine side, but for now, this is what I have done to accomplish this on a production environment (the Skytap env I have up right now).  This shouldn't affect your local build as Alpine is started through sbt there, but just a heads up that that is still a thing to consider.

kevinmtrowbridge:
@jleealpine: thank you!  I will be circling back to this shortly.

PR #293, opened 2016-04-14
DEV-14471: Tags banished from the project cards

imccarter:
Affected places: project card template + some styles, succinct response from workspace presenter, and project card view.


imccarter:
rake tmp:cache:clear is necessary to "reset" the presenter to not respond with tags for the card's workspace.

mmilano:
@imccarter i think that workspace_presenter_spec.rb 
may also need to be re-verted where it checks for the tags in the response

mmilano:
other than the re-adjustment noted above (and presuming there are no other specs), should be good to go.,

PR #296, opened 2016-04-20
DEV-14471: Remove unnecessary function

imccarter:
DEV-14471: The dashboard project list view had a function that was a workaround to display the tags on each project card. It was getting called on every 'reset' of the project card list, but did nothing now that the tags aren't shown. This PR removes the function, further cleaning out anything tied to displaying tags on the project card.


mmilano:
ok.

PR #299, opened 2016-04-21
MONITORING 6.0 supervisord -- 2nd time's the charm

kevinmtrowbridge:
@jleealpine and I will collaborate around this new PR to try again to get the "upgrading" case for the installer working with supervisord.

We'll hopefully merge this in early next week.  Apologies.  Thank you for your patience.


kevinmtrowbridge:
This came a long way today:
* I'm able to upgrade from 5.9.1 to 6.0 now.
* I restyled the status control UI to match Alpine's brand:
alpine status

I need @haozou to review this.  I will comment on some places I need you to look at @haozou.

@nkapinos: This branch has been packaged and has an installer here: http://10.10.0.215:8080/job/alpine-chorus-packaging-branch-monitoring/ ... This is fair game for testing.  I will discuss with you tomorrow.

components/core/app/services/service_scheduler.rb
```diff
@@ -37,11 +37,6 @@ def initialize
       SolrIndexer.SolrQC.enqueue_if_not_queued('SolrIndexer.refresh_external_data')
     end
 
-    every(ChorusConfig.instance['restart_indexer_interval_hours'].hours, 'chorus_control.sh restart indexer') {
-      chorus_control = File.expand_path("../../../packaging/chorus_control.sh", __FILE__)
-      `#{chorus_control} restart indexer`
-    }
-
```
kevinmtrowbridge: @midnighteuler: I think I am going to strip this out, since the whole point of this supervisord is to provide a better way to solve this problem.  This is cause for celebration, what we did is super ugly.

docker/build/mods/packaging/chorus_control.sh
```diff
@@ -0,0 +1,373 @@
...
+  if should_handle postgres;  then
+    $CHORUS_HOME/bin/supervisorctl start postgres
+    EXIT_STATUS=`expr $EXIT_STATUS + $?`;
+  fi
+  if should_handle solr;      then
+    if ! contains ${services[@]} solr && [[ $INDEXER_DISABLED == 1 ]]; then
+        log "Skipping solr (See indexer.enabled in chorus.properties)...";
+    else
+        STARTING="solr"
+        # depends_on postgres
```
kevinmtrowbridge: Folks can mostly ignore this, the point of this is just to modify chorus_control.sh to work with the Travis CI Docker test running.

packaging/build_scripts/build-postgres.sh
```diff
@@ -18,7 +18,7 @@ pushd /tmp
 echo "Packaging postgres-$POSTGRES_VERSION.tar.gz.."
 tar czfh postgres-$POSTGRES_VERSION.tar.gz postgres/
 
-echo "Moving package to $CHORUS_HOME/packaging/postgres/.."
+echo "Moving package to $CHORUS_HOME/vendor/postgres/.."
```
kevinmtrowbridge: Ok, this is just the postgres build scripts which only devs will use infrequently -- but this message is inaccurate, it doesn't match the command above.

packaging/chorus-config-supervisor.sh
```diff
@@ -0,0 +1,86 @@
...
+export MIZUNO_PID_FILE=$MIZUNO_PID_FILE
+export JETTY_PID_FILE=$JETTY_PID_FILE
+export ALPINE_PID_FILE=$ALPINE_HOME/alpine.pid
+
+export SOLR_PID_FILE=$SOLR_PID_FILE
+export NGINX_PID_FILE=$NGINX_PID_FILE
+export SCHEDULER_PID_FILE=$SCHEDULER_PID_FILE
+export WORKER_PID_FILE=$WORKER_PID_FILE
+export INDEXER_PID_FILE=$INDEXER_PID_FILE
+export SUPERVISORD_PID_FILE=$SUPERVISORD_PID_FILE
```
kevinmtrowbridge: @haozou: the way @jleealpine and I came up with, removing all the service-specific config of these ENV vars (from the `start-x.sh` files that we deleted), was to move all the ENV vars into this script, and to run it before the supervisord daemon is configured.

Here is the supervisord documentation that describes how the ENV vars defined at the time of starting the supervisord daemon are preserved and reused when reissuing the start commands:

> Environment Variables
> Environment variables that are present in the environment at the time that supervisord is started can be used in the configuration file using the Python string expression syntax %(ENV_X)s
http://supervisord.org/configuration.html#environment-variables

packaging/chorus-config-supervisor.sh
```diff
@@ -0,0 +1,86 @@
...
+  if [ -f "$script_dir_path"/supervisord/programs/alpine.conf ];
+  then
+    rm "$script_dir_path"/supervisord/programs/alpine.conf
+  fi
+}
+
+if [ -f $ALPINE_HOME/scripts/start_alpine.sh ];
+then
+    enable_alpine
+    # . $ALPINE_HOME/scripts/start_alpine.sh prepare
```
kevinmtrowbridge: @jleealpine: the `chorus-config-supervisor.sh` script was actually getting called at a seemingly innocuous location ... and calling into this point, and it was actually trying to start alpine (ie, the "prepare" part wasn't working) ... and, so it didn't have enough memory and threw a strange error and, it was a little confusing.

(This is from line 1418 of the install.log, when upgrading from 5.9 -> 6.0 -- note how it's actually trying to `/usr/local/chorus/chorus_control.sh stop` Ha!):
```
827a4bc77/postgres/bin:$PATH && CHORUS_HOME=/usr/local/chorus/current ALPINE_HOME=/usr/local/chorus/alpine-current ALPINE_DATA_REPOSITORY=/usr/local/chorus/shared/ALPINE_DATA_REPOSITORY /usr/local/chorus/chorus_control.sh stop
2016-04-21 20:23:13,525 - chorus - DEBUG : /data/chorus/tmp
Starting Jetty: Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x000000066aaa0000, 2863333376, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 2863333376 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /usr/local/chorus/alpine-releases/alpine-5.9.1.0-36-d4424c7b6/jetty-distribution-9.2.12.v20150709/hs_err_pid3292.log
OK Thu Apr 21 20:23:13 UTC 2016
[production] Supervisor: all processes stopped, shutting down supervisor daemon.

2016-04-21 20:23:13,525 - chorus - DEBUG : /usr/local/chorus/releases/6.0.0.0.4101-827a4bc77/packaging/stop-
```

So I commented it out -- and Alpine seems to be starting anyways??  We will need to follow up on this early next week.

packaging/chorus-config-supervisor.sh
```diff
@@ -0,0 +1,86 @@
...
+else
+    disable_alpine
+fi
+
+
+## Enable or disable Jetty / Mizuno based on the $RAILS_ENV
+
```
kevinmtrowbridge: @haozou: note how this section works -- either running mizuno or jetty, and starting Alpine or, not starting alpine, based on checking for the presence of files and, either copying certain *.conf files into the `packaging/supervisord/programs` directory, or deleting them.

If there were an easier way to do this (like maybe with ENV vars?) I would love to do that, but, this doesn't exist currently -- for proof, see this feature request in the Supervisord Github issue tracker: https://github.com/Supervisor/supervisor/issues/591

packaging/chorus-config-supervisor.sh
```diff
@@ -0,0 +1,86 @@
...
+    cp "$script_dir_path"/supervisord/programs/templates/jetty.conf.template "$script_dir_path"/supervisord/programs/jetty.conf
+  fi
+  if [ -f "$script_dir_path"/supervisord/programs/mizuno.conf ];
+  then
+    rm "$script_dir_path"/supervisord/programs/mizuno.conf
+  fi
+fi
+
+
+export PYTHONPATH="$script_dir_path"/../vendor/python_lib/lib/python/:$PYTHONPATH
```
kevinmtrowbridge: @haozou: note this humble spot is where the PYTHONPATH is configured to point at the `vendor/python_lib` directory ... this won't wear well if we install other python libs that don't have to do with supervisor, but I wasn't sure what a better spot might be, since we don't actually copy in any of the Python installation files, until we run the packaging scripts.

Do you know of a better place?

packaging/chorus-config.sh
```diff
@@ -58,6 +60,7 @@ SCHEDULER_PID_FILE=$CHORUS_HOME/tmp/pids/scheduler.$RAILS_ENV.pid
 WORKER_PID_FILE=$CHORUS_HOME/tmp/pids/worker.$RAILS_ENV.pid
 INDEXER_PID_FILE=$CHORUS_HOME/tmp/pids/indexer.$RAILS_ENV.pid
 MIZUNO_PID_FILE=$CHORUS_HOME/tmp/pids/mizuno.pid
```
kevinmtrowbridge: I actually think we need to cleanup / delete all of these -- we aren't tracking the processes via PID files anymore.

packaging/chorus-config.sh
```diff
@@ -131,33 +134,6 @@ function wait_for_stop () {
     echo " ( Stopped )"
 }
 
-DEFAULT_WAIT_TIME=5
-function wait_for_stop_or_force () {
-    pid_file=$1
```
kevinmtrowbridge: @haozou: I deleted all this because, supervisord implements this natively.  See `stopwaitsecs` here in the documentation: http://supervisord.org/configuration.html#program-x-section-values

packaging/chorus_control.sh
```diff
@@ -1,12 +1,13 @@
...
+if [ "$script_dir_path" == "" ]; then
+ script_dir_path=$0
 fi
-bin=`dirname "$bin"`
-bin=`cd "$bin"; pwd`
+script_dir_path=`dirname "$script_dir_path"`
+script_dir_path=`cd "$script_dir_path"; pwd`
 
-. "$bin"/chorus-config.sh
+. "$script_dir_path"/chorus-config.sh
```
kevinmtrowbridge: Hi @haozou: renaming this isn't really necessary, I just found it hard to understand that all this is just a way of getting the absolute path to the directory where the script is located ... `$bin` wasn't communicating that to me, so I renamed it, but I didn't do it consistently, so we should either change it back or, change the other places where we use `$bin` still in these scripts.

packaging/chorus_control.sh
```diff
@@ -51,17 +52,34 @@ function configure () {
...
 }
 function start () {
+
+  . "$script_dir_path"/chorus-config-supervisor.sh
+
+  SUPERVISORD_PID_FILE=$CHORUS_HOME/tmp/pids/supervisord.pid
+  if ( test -f $SUPERVISORD_PID_FILE ) && ( kill -0 `cat $SUPERVISORD_PID_FILE` > /dev/null 2>&1 ); then
+      log "supervisord already running as process `cat $SUPERVISORD_PID_FILE`."
+      log "rereading supervisord cofiguration ..."
+      "$script_dir_path"/../bin/supervisorctl reread
```
kevinmtrowbridge: `supervisorctl reread` also causes the ENV vars that supervisord is storing, to be re-evaluated, so if someone edits say, the `chorus.properties` file, and restarts, their edits will be drawn in.  (I think, we should test this.)

packaging/chorus_control.sh
```diff
@@ -51,17 +52,34 @@ function configure () {
...
+
+      wait_for_start $SUPERVISORD_PID_FILE
+      log "supervisord started as pid `cat $SUPERVISORD_PID_FILE`"
+  fi
+
   EXIT_STATUS=0
   pushd $CHORUS_HOME > /dev/null
   if should_handle postgres;  then
-    $bin/start-postgres.sh;
+    $CHORUS_HOME/bin/supervisorctl start postgres
```
kevinmtrowbridge: I guess we may be able to use `$CHORUS_HOME` here instead of this gross looking `"$script_dir_path"/../bin` ... earlier I thought the scripts were failing because of a path issue, but then I came later to understand more about the sequencing of the re-linking of the release-dir into current, and the updating of the link to `chorus_control.sh` in the install directory structure.

We should standardize on one or the other of these ways.

packaging/chorus_control.sh
```diff
@@ -70,25 +88,35 @@ function start () {
     if ! contains ${services[@]} indexer && [[ $INDEXER_DISABLED -eq 1 ]]; then
         log "Skipping indexer (See indexer.enabled in chorus.properties)...";
     else
-        $bin/start-indexer.sh;
+        STARTING="indexer"
+        depends_on postgres
```
kevinmtrowbridge: There is no way to handle this `depends_on` functionality in supervisord, so we need to preserve this.  I tried to delete it.  See: https://github.com/Supervisor/supervisor/issues/122

packaging/chorus_control.sh
```diff
@@ -105,7 +133,7 @@ function migrate() {
   if [[ "$@" == "-f" || "$@" == "force" ]]; then
     flag=true;
   fi
-  $bin/start-postgres.sh;
+  "$script_dir_path"/start-postgres-simply.sh;
```
kevinmtrowbridge: @haozou: can you help me with this?

Postgres is the one dependency that must be started during the install process, and at this point, imagine that we had an already running Chorus 6.2 installation, and we were upgrading to 6.3 ... so we tried to start Postgres using supervisord -- doesn't it seem like this would conflict?  

That is why I preserved this other way of starting postgres, so there are two right now: `start-postgres-simply.sh` and `supervisorctl start postgres` ...

But -- I'm not sure about the interaction of different postgres instances when upgrades are being run ....

For example, at line 1400 of the chorus 5.9 -> 6.0 upgrade install.log, I see:

```
2016-04-22 03:20:12,751 - chorus - DEBUG : ->Running data validation...
2016-04-22 03:20:12,752 - chorus - DEBUG : Starting postgres...
2016-04-22 03:20:12,752 - chorus - DEBUG : PATH=/usr/local/chorus/releases/6.0.0.0.6-5e6aeb37c/postgres/bin:$PATH && CHORUS_HOME=/usr/local/chorus/releases/6.0.0.0.6-5e6aeb37c ALPINE_HOME=/usr/local/chorus/alpine-current ALPINE_DATA_REPOSITORY=/usr/local/chorus/shared/ALPINE_DATA_REPOSITORY /usr/local/chorus/releases/6.0.0.0.6-5e6aeb37c/packaging/start-postgres-simply.sh
2016-04-22 03:20:12,804 - chorus - DEBUG : [production] Postgres already running as process 2275.

2016-04-22 03:20:12,805 - chorus - DEBUG : PATH=/usr/local/chorus/releases/6.0.0.0.6-5e6aeb37c/postgres/bin:$PATH && cd /usr/local/chorus/releases/6.0.0.0.6-5e6aeb37c && RAILS_ENV=production bin/ruby -S bin/rake validations:data_source --trace
```

Note how it says: `Postgres already running as process 2275.` ... I guess I'm not totally clear on what the whole `->Running data validation...` step is for.

packaging/chorus_control.sh
```diff
@@ -127,53 +158,47 @@ function stop () {
...
+    "$script_dir_path"/../bin/supervisorctl shutdown
+  fi
+
   popd > /dev/null
   if (($EXIT_STATUS > 0)); then
     exit_control $EXIT_STATUS;
   fi
 }
 
-function monitor () {
```
kevinmtrowbridge: I scrubbed this stuff out, cause a more sophisticated monitoring is what Supervisord is for ...

packaging/chorus_control.sh
```diff
@@ -278,13 +303,12 @@ function run_in_root_dir_with_postgres () {
...
 function usage () {
   script=`basename $0`
-  echo "$script is a utility to start, stop, restart, or monitor the Chorus services."
+  echo "$script is a utility to start, stop, or restart Chorus services."
   echo
   echo Usage:
   echo "  $script start   [services]                       start services"
-  echo "  $script stop    [-t max_wait_time] [services]    stop services"
-  echo "  $script restart [-t max_wait_time] [services]    stop and start services"
-  echo "  $script monitor [services]                       monitor and restart services as needed"
```
kevinmtrowbridge: This command flag `-t` has gone away as previously mentioned.

packaging/setup/chorus_setup.py
```diff
@@ -61,12 +61,12 @@ def construct_shared_structure(self):
...
         self._mkdir_p(os.path.join(self.shared, "solr"))
         self._mkdir_p(os.path.join(self.shared, "tmp"))
-        self._cp_if_not_exist(os.path.join(self.release_path, "packaging/database.yml.example"), \
+        self._cp_if_not_exist(os.path.join(self.release_path, "packaging/templates/database.yml.example"), \
                               os.path.join(self.shared, "database.yml"))
         self._cp_if_not_exist(os.path.join(self.release_path, "config/chorus.defaults.properties"), \
                               os.path.join(self.shared, "chorus.properties"))
         os.chmod(os.path.join(self.shared, "chorus.properties"), 0600)
-        self._cp_f(os.path.join(self.release_path, "packaging/sunspot.yml.example"), \
+        self._cp_f(os.path.join(self.release_path, "packaging/templates/sunspot.yml.example"), \
```
kevinmtrowbridge: This was just some refactoring to cut down on the big num of files in the `packaging` directory.

packaging/setup/chorus_setup.py
```diff
@@ -138,12 +138,14 @@ def link_data_folder(self):
         self._ln_sf(os.path.join(self.shared, "log/nginx"), \
                    os.path.join(self.release_path, "vendor/nginx/nginx_dist/nginx_data/logs"))
 
-        self._ln_sf(os.path.join(self.release_path, "packaging/chorus_control.sh"), \
-                   os.path.join(self.options.chorus_path, "chorus_control.sh"))
         self._ln_sf(os.path.join(self.release_path, "packaging/setup/chorus_server"), \
```
kevinmtrowbridge: @haozou: I especially want to run this by you -- I found that, the `chorus_control.sh` script was being updated too soon, so that, when the installer says `chorus - DEBUG : ->Shutting down previous Chorus install...`, it was actually running the NEW `chorus_control.sh` script to do it.  This was failing, because I had changed the script so much.  It wasn't able to shut down the old instance using the new code.

packaging/setup/chorus_setup.py
```diff
@@ -457,6 +459,7 @@ def setup(self, options, is_upgrade):
         self.generate_chorus_rails_console_file()
 
         #self.clean_up_old_releases()
+        self.link_chorus_control()
```
kevinmtrowbridge: So I moved the relinking of `chorus_control.sh` further down in the process.

packaging/start-postgres-simply.sh
```diff
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
+
+# Starting and stopping these services is controlled by supervisord.  Except postgres is needed by
+# the installer, so here there's an additional way to start & stop it, used by the installer.
+
```
kevinmtrowbridge: My explanation for why this `start-postgres-simply.sh` exists, this file might not really be necessary? (See previous comments.)

packaging/start-webserver.sh
```diff
@@ -15,50 +14,15 @@ bin=`cd "$bin"; pwd`
...
-    cd $CHORUS_HOME/vendor/nginx/nginx_dist/
-
-    OPENSSL_CONF=$OPENSSL_CONF ./$NGINX
-    wait_for_start $NGINX_PID_FILE
-    log "nginx started as pid `cat $NGINX_PID_FILE`"
-fi
-
-cd $CHORUS_HOME
+RAILS_ENV=$RAILS_ENV CHORUS_HOME=$CHORUS_HOME $RUBY vendor/nginx/generate_nginx_conf.rb
+"$script_dir_path"/../bin/supervisorctl start nginx
```
kevinmtrowbridge: @jleealpine and I thought it worth keeping this file just because it does a little more than the others, it starts nginx, runs `generate_nginx_conf` ... and also has an `if` statement to either run jetty if in production and mizuno if in dev.  

(As a side note, we're actually thinking we can maybe get rid of mizuno and just use webbrick?  @midnighteuler and I aren't sure if Mizuno is necessary.)

packaging/start-webserver.sh
```diff
@@ -15,50 +14,15 @@ bin=`cd "$bin"; pwd`
...
-            $RUBY packaging/update_jetty_xml.rb
-            RAILS_ENV=$RAILS_ENV $RUBY packaging/update_database_yml.rb
-
-            log "starting jetty..."
-            cd $CHORUS_HOME/vendor/jetty/
-            JETTY_PID=$JETTY_PID_FILE CHORUS_JAVA_OPTIONS=$CHORUS_JAVA_OPTIONS RAILS_ENV=$RAILS_ENV SOLR_PORT=$SOLR_PORT ./jetty-init start 2>> $CHORUS_HOME/log/jetty.log >> $CHORUS_HOME/log/jetty.log &
-            cd $CHORUS_HOME
-            wait_for_start $JETTY_PID_FILE
-			log "jetty started as pid `cat $JETTY_PID_FILE`"
-        fi
```
kevinmtrowbridge: The coalescing of all this crud into a single line is one spot where this Supervisord feature really shines.

packaging/supervisord/programs/indexer.conf
```diff
@@ -0,0 +1,14 @@
...
+environment=QUEUE="indexer_queue",JRUBY_OPTS="%(ENV_JRUBY_OPTS)s",CHORUS_JAVA_OPTIONS="%(ENV_CHORUS_JAVA_OPTIONS_WITHOUT_XMS)s",RAILS_ENV="%(ENV_RAILS_ENV)s",SOLR_PORT="%(ENV_SOLR_PORT)s"
+redirect_stderr=true
+stdout_logfile=%(ENV_CHORUS_HOME)s/log/indexer.log
+stopwaitsecs=30
+
+# seems to be necessary
+stopsignal=QUIT
+
+# important
+priority=600
```
kevinmtrowbridge: These files are kinda awesome.  There are more options we can pass in here: http://supervisord.org/configuration.html#program-x-section-values

`stopsignal` should maybe be named `initialstopsignal` because, Supervisord will hard KILL the process after the `stopwaitsecs` period has passed.  It tries sending the `stopsignal` every few seconds, until then.

We also define the spot where the STDOUT logs are written to disk here.  And we tell it to put down STDERR and STDOUT together, into a single log file.

Also, we tell it not to `autostart`, because we are letting `chorus_control.sh` handle that.  But, we set `autorestart` to true.  autorestart might be something we would want to turn off in dev?

packaging/supervisord/programs/postgres.conf
```diff
@@ -0,0 +1,15 @@
...
+autostart=false
+command=%(ENV_CHORUS_HOME)s/postgres/bin/postgres -D %(ENV_POSTGRES_DATA_DIR)s -N%(ENV_MAX_CONNECTIONS)s -p%(ENV_POSTGRES_PORT)s -h127.0.0.1 --bytea_output=escape
+environment=DYLD_LIBRARY_PATH=%(ENV_CHORUS_HOME)s/postgres/lib,LD_LIBRARY_PATH=%(ENV_CHORUS_HOME)s/postgres/lib
+redirect_stderr=true
+stdout_logfile=%(ENV_CHORUS_HOME)s/log/postgres.log
+stopasgroup=true
+stopwaitsecs=30
+
+# important
+priority=1
```
kevinmtrowbridge: Postgres is the most important service and I set it to be the first to start, and last to stop.  But this `priority` setting is not quite what it's all cracked up to be: https://github.com/Supervisor/supervisor/issues/122

packaging/supervisord/supervisord.conf
```diff
@@ -0,0 +1,22 @@
+; http://supervisord.org/index.html
+
+[unix_http_server]
+file=/tmp/supervisor.sock   ; (the path to the socket file)
+
+[inet_http_server]         ; inet (TCP) server disabled by default
+port=:9001                 ; (ip_address:port specifier, *:port for all iface)
```
kevinmtrowbridge: This is where the config for the Web UI panel goes.  Note we can configure it to require a password (which maybe we could auto-POST from the Admin panel) to put a little more security in place, cause now, anyone can see the control panel: http://supervisord.org/configuration.html#inet-http-server-section-settings

I'm sure we will want to do this.

packaging/supervisord/supervisord.conf
```diff
@@ -0,0 +1,22 @@
...
+
+[supervisord]
+pidfile=%(ENV_CHORUS_HOME)s/tmp/pids/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
+childlogdir=%(ENV_CHORUS_HOME)s/log
+logfile=%(ENV_CHORUS_HOME)s/log/supervisord.log ; (main log file;default $CWD/supervisord.log)
+loglevel=info             ; (log level;default info; others: debug,warn,trace)
```
kevinmtrowbridge: Set supervisord daemon's log level, useful for debugging this whole setup.

packaging/supervisord/supervisord.conf
```diff
@@ -0,0 +1,22 @@
...
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisorctl]
+serverurl=unix:///tmp/supervisor.sock ; use a unix:// URL  for a unix socket
+
+[include]
+files=./programs/*.conf
```
kevinmtrowbridge: Again this is how the conditional enabling / disabling of jetty, mizuno, alpine is working, by creating or deleting their process specific conf file in this directory.

vendor/python_lib/lib/python/easy-install.pth
```diff
@@ -0,0 +1,5 @@
+import sys; sys.__plen = len(sys.path)
```
kevinmtrowbridge: So all this can be mostly ignored, it's just the source for supervisord and it's dependencies.

PR #306, opened 2016-04-28
DEV-14548 Workspace membership

imccarter:
Standardized "badge" (... or, "card"?) for workspace members that includes avatar, name, and workspace role. New layout for the workspace member module that appears on the workspace list page when a single workspace is selected and on the workspace show page. Button to "View All" when there are more than 7 total members will open a modal displaying the same "badges" in two columns that can be sorted by first or last name.


components/frontend/app/assets/javascripts/views/workspaces/workspace_member_list_view.js
```diff
@@ -8,6 +8,7 @@ chorus.views.WorkspaceMemberList = chorus.views.Base.extend({
 
     setup: function() {
         this.subscribePageEvent("workspace:selected", this.setWorkspace);
+        this.maxMemberCount = 6;// how many members are displayed on the sidebar
```
mmilano: can you move this to the top level of the view object, so that it is listed at the top of the object definition (e.g. by templatename)?
maxMemberCount: 6;

it will be more readable (to other people) and more accessible (in case we use it elsewhere) that way,
and anyway, this scoping is already doing that.

imccarter: Good catch, done.

mmilano:
lg.

PR #308, opened 2016-04-29
DEV-14548 Fix some tests for workspace member sidebar

imccarter:
I overlooked double-checking the jasmine tests, but fortunately it was an easy fix to get them working with the changes.


components/frontend/Gemfile.lock
```diff
@@ -21,6 +21,29 @@ GIT
       rspec
 
 PATH
```
imccarter: I'm not sure why .gitignore isn't actually ignoring this...

components/frontend/spec/javascripts/pages/dashboard_page_spec.js
```diff
@@ -8,7 +8,7 @@ describe("chorus.pages.DashboardPage", function() {
 
         afterEach(function() {
             this.page = null;
-        };
+        });
```
imccarter: No clue when/where this happened, but I just happened to catch it randomly.

PR #309, opened 2016-05-02
Dataset remove association

mmilano:
fixes IBX-3507, IBX-3985, IBX-1345

changes the actions shown to users for disassociating datasets from a workspace
from 'delete association' to either remove or delete (as appropriate).

all datasets are 'removed' (not deleted) except for chorus-views, which are actually deleted.
changes are continued through to the dialogs as well.


mmilano:
@imccarter could you take a look through this please?

components/frontend/app/assets/javascripts/alerts/dataset_delete_alert.js
```diff
@@ -4,11 +4,30 @@ chorus.alerts.DatasetDelete = chorus.alerts.ModelDelete.extend({
...
+        // except: chorus_view is an entity that only exists in a workspace, and so it is actually deleted
+        switch(keyType) {
+            case "chorus_view":
+                keyPrefix = "dataset_delete.";
+                break;
+            case "table":
+            case "view":
+            case "hdfs_dataset":
+                keyPrefix = "dataset_remove.";
+                break;
```
imccarter: I think lines 17-21 are unnecessary if you want those cases to use the default behavior. Though, 17-19 are probably good for clarity and possible changes in the future.

components/frontend/spec/javascripts/alerts/dataset_disassociate_multiple_alert_spec.js
```diff
@@ -1,4 +1,7 @@
 describe("chorus.alerts.DatasetDisassociateMultiple", function() {
+    var collection;
+    var alert;
```
imccarter: Are these variables used?

imccarter:
lgtm

components/frontend/app/assets/javascripts/alerts/dataset_delete_alert.js
mmilano: i prefer to err here on the side of greater clarity and future flexibility.

PR #310, opened 2016-05-02
DEV-14625 / DEV-14689 workspace sidebar visual bugs

imccarter:
As noted in the Jira comments; I also took the opportunity to get rid of the "Activity" lozenge box that unnecessarily rendered on initial page load (DEV-14689). I'm not 100% satisfied with using jquery hide() and show(), but can't think of a better way without getting too crazy; open to suggestions.


mmilano:
@imccarter 
if i read this correctly, it appears to call js to hide the content in the js on setup.

wouldnt it be better to have the tabs hidden by default in the tamplate & via css,
and then make sure they are visible when there is a selection?

imccarter:
Yes, that is the case. I might be missing/forgetting something, but won't js have to get involved somewhere as to show or hide the content on (de)selection? Either via class assignment or like what I have here? 

Hiding them as a default via css and then un-hiding on selection (via js?) would probably be preferable and I can definitely do that. I'll play around, I could be overlooking some obvious fix...

imccarter:
@mmilano 
Sorry for the delay. So the `$el.show()`s are still in there to have it display when a workspace is selected, but I've turned the `$el.hide()`s into `display: none` so it just initializes as hidden.

imccarter:
Just noting that there's an issue for the 'Activity' box now; DEV-14689

PR #311, opened 2016-05-02
DEV-14367 Make button text in sidebar uniformly grey

imccarter:
The "View All" button for the workspace membership module and the "More" button for the activity sidebar (on the workspace list view) should match the coloring for the "More" button on the main activity feed on the workspace show page.


mmilano:
LGTM

PR #313, opened 2016-05-03
DEV-14658 hide grey list footer while loading

imccarter:
An admittedly somewhat hacky fix to get the list content footer to not show while the collection is being fetched. It would also inappropriately be visible while loading a list that wouldn't display it.


imccarter:
Hold off on merging this, I'm going to change it so that it uses some class assignment rather than show to get the same effect. Still not the perfectly ideal fix, but a bit cleaner.

imccarter:
As per Josh's recommendation I switched to using classes to hide then show the footer rather than using show(). I also made the listener for this a conditional `listenToOnce` for a bit cleaner implementation and changed the name of the option flag for clarity.

mmilano:
@imccarter github tells me that this has conflicts and needs to be resolved.
however, is the problem still a problem? was this already merged? i havent been seeing the problem on pages...

imccarter:
@mmilano I'm going to force push a fresh branch of these fixes here; merging master had some nasty conflicts that borked everything post-workspace list rework.

PR #315, opened 2016-05-04
DEV-14653 Pass workspace id to tag, not workfile id

imccarter:
Clicking on workfile tags would link to the tag's show page associated with a random workspace or bring up an error as it was using the associated workfile id, not the workspace id; this fixes that.


jmlewis:
LGTM

PR #318, opened 2016-05-04
DEV-14673 Long owner names in workspace list items have a max-width

imccarter:
I also increased the general width of both the workspace stage badge and owner name column. Just a little more room so that realistically long names aren't cut off. This may need further refinement if/when the owner column is aligned left.


mmilano:
incorporated some of these changes and resolved this issue in larger commit
https://github.com/Chorus/chorus/commit/11eee90478b9b97fed92c269ac6814fd590d9987

mmilano:
closed. can be deleted.

PR #319, opened 2016-05-05
DEV-14667 Add job item specific rules for loading spinner

imccarter:
CSS rules that overwrite the more general .loading_spinner rules for job items so it renders in the correct position.


imccarter:
As I'm moving _job_list.scss to where it should be in the stylesheets/jobs folder, I'll be combining this PR with DEV-14716

PR #321, opened 2016-05-06
DEV-14682 Case-insensitive sorting by name for workspace members

imccarter:
The workspace members were being displayed sorted case-sensitive (lower-case appearing after any upper-case) in the "View All" modal, while the sidebar had no discernible sorting in place. However, the add/edit members dialog sorted them case-insensitive. This fix makes all sort functions for workspace members case-insensitive, and adds this sorting to the sidebar (by first name, then last).


jmlewis:
LGTM

PR #322, opened 2016-05-07
DEV-14594 Separate the workspace index vs show page behavior for membership widget

imccarter:
There was some difference in the way the membership side should've been handled on the workspace show page vs. workspace index when one is selected. Not accounting for this resulted in a bad api call; this fixes it. Also commented out a superfluous render(), as the membership widget renders on page load but is hidden initally, and is displayed with show(). This is a workaround, and is something that should be looked into down the road; ideally it shouldn't exist until a workspace is selected and then render should be called.


jmlewis:
LGTM

PR #323, opened 2016-05-09
DEV-14717 Fix workspace owner name wrapping

imccarter:
Some css to handle long owner names in the workspace membership module.


components/frontend/app/assets/stylesheets/workspaces/_workspace_sidebar.scss
```diff
@@ -46,6 +46,14 @@
 
             .owner_link {
                 padding-bottom: 5px;
+
+                a {
```
mmilano: there is a mixin defined in _common named ellipsis
that adds the css pieces necessary here;
better to use that (common usage, etc).

imccarter: Done and pushed.

PR #325, opened 2016-05-10
DEV-14268 workspace list filter

imccarter:
The stages are stored as "stage1", "stage2", etc. in the db; getting the search to hit those properly will be pushed back to a later release. This PR allows the search to work against the workspace name and workspace owner's username or first/last name.


components/core/app/models/workspace.rb
```diff
@@ -92,7 +92,7 @@ def self.eager_load_associations
      :archiver,
      :tags,
      :owner,
-     {:sandbox => {:scoped_parent => {:data_source => [:tags, {:owner => :tags}]}}}
+     :sandbox
```
imccarter: This is the change I'm most concerned about; I'm taking out code added ~2013 or earlier, and I'm not sure what further ramifications that might have.

imccarter:
@jleealpine, Michel suggested having you review this PR; I've left a comment on a specific change I made that we could use another pair of eyes on. Otherwise it's pretty similar to the search feature for users that you put in.

components/core/app/models/workspace.rb
jleealpine: I'm not entirely sure what this is about either, is there a specific reason that you're removing it?

imccarter: Without removing this: when I use .references(Workspace.eager_load_associations) in the workspaces controller, I'd get an Active Record "Cannot eagerly load polymorphic association ':scoped_parent'" error. I need to use that reference to search against the users (owner) table.

jleealpine:
LGTM

PR #326, opened 2016-05-11
Fix a test, refactor workspace sidebar rendering

imccarter:
Clean up a workaround for the rendering bugs in the workspace index sidebar. Now the membership module and activity tab won't render at all until a workspace is selected, rather than initially hiding them and showing them on selection. This workaround also broke a test which is now fixed with some slight adjustment.


jmlewis:
LGTM

PR #327, opened 2016-05-11
DEV-14658 Hide the pagination list footer while collection is being fetched

imccarter:
Master currently just leaves the footer "hidden", but it's there. This PR in effect adds the pagination stuff to the bottom of the list where necessary, and fixes the bug DEV-14658 references that was part of the first implementation of the footer.


jmlewis:
LGTM

PR #328, opened 2016-05-12
DEV-14275 Add an exception for the summary on workspace search results

imccarter:
The description of the workspace will now only appear as a fragment of where the search query is, and not display otherwise. On selecting a workspace, the entire description (possibly truncated) will appear on the sidebar.

As far as I can tell; the "summary" attribute is only in workspaces so this change should have no effect on any other search result item.


mmilano:
looks fine

PR #330, opened 2016-05-12
DEV-14830 Fix some overflowing tags

imccarter:
Fix overflowing tags on user search results. They will now just wrap onto another line. This will possibly increase the width of the item, but I figure all the tags should display in search results, as they can be searched against. Also add a temp fix for a small visual bug on user avatar in search results.


jleealpine:
LGTM

PR #331, opened 2016-05-12
DEV-14808 Update workspace same-stage fix

imccarter:
Button will be disabled until a stage different from the current stage is selected.


jmlewis:
LGTM

PR #332, opened 2016-05-13
DEV-14855 Workspace action permissions

imccarter:
In fixing DEV-14855, it became apparent that there were no differences in the actions members vs non-members can make from the workspace action sidebar. This PR essentially sets up three levels of permission; non-members, members, and owner/admin. Actions available to non-members are the same actions available to anyone from the workspace list; namely 'add insight', 'add/edit tags', and 'add note'. The rest of the actions are available to members, except 'delete workspace' and 'add/edit members', which are only available to workspace owners or a general admin. Moving any of these actions around should be pretty easy now, too.


imccarter:
Ok, this broke a handful of tests. Will address those tomorrow and add to this PR.

imccarter:
This should be good to go. Possible room for debate in what a non-member can do with a public workspace: with this implementation, a non-member can add a note, add an insight, or add/edit tags (which are also the available actions for everyone from the workspace index page). Will be easy to adjust now if we decide otherwise.

jmlewis:
LGTM

PR #333, opened 2016-05-13
DEV-14825 Admin users who are user_type = Collaborator, cannot open workflows

kevinmtrowbridge:
* DEV-14825 A user with "admin" permissions has full permissions to a python notebook (unlike workflows), in a workspace where they are not a member

I think the real bug here is that Admin users who are collaborators cannot open workflows.  In this PR, I add the 'create_workflow' permission, if the user is an admin.  Please see the full discussion at:
https://alpine.atlassian.net/browse/DEV-14825


kevinmtrowbridge:
Per our meeting this morning, this change is not necessary: https://alpine.atlassian.net/browse/DEV-14825

So -- closing this.

PR #334, opened 2016-05-13
DEV-14904: hide comment link for non-members who view public workspaces

kevinmtrowbridge:
This is a potential solution for:

* DEV-14904 User Can't post comment to a PUBLIC workspace which user is NOT a member of.

This hides the "Comment" link for users who are not members of workspaces, who view those workspaces (must be public workspaces?).

Assuming that non-members should not be able to comment on public workspaces.  I'm not sure, this seemed like the best solution, so I went ahead and developed it.


imccarter:
As a note; I'm working on permissions for the workspace action panel sidebar now, and we're letting non-members add insights, tags, and notes to public workspaces. Comments might be worth considering.

kevinmtrowbridge:
https://alpine.atlassian.net/browse/DEV-14904
Per conversation this morning in the engineering triaging meeting, I merged this in.

PR #335, opened 2016-05-16
DEV-14936 Change workspace owner badge to use properly scaled icon-size avatar

imccarter:
Leaving the owner-badge avatar that displays in the workspace show page sidebar (membership module) as is because it's a bit larger than the user-list page avatar.


jmlewis:
LGTM

PR #339, opened 2016-05-16
move error serialization into the API component

kevinmtrowbridge:
Move error serialization into the Api component, so that we can use the traditional Rails validation code in other components.

The specific reason why this is necessary is, when I was building the Email component, where I am using the traditional Rails error messaging, I noticed that the error messages were coming through formatted for the API.

errors-1

As we aim to move more stuff out of the chorus.properties file, and into feature specific panes within the Admin console, we will want to build out a lot of form validation in this way.

So, this goal of this PR is to put the API specific JSON error serialization, in the Api only, and allow the ActiveRecord models in Core, to be interacted with in a normal way, by other components, like Admin.


kevinmtrowbridge:
This was merged into #278, so, it's no longer necessary to have an independent PR.  Closing.

PR #343, opened 2016-05-23
DEV-14969 Switch from _map to reduce() to bypass deleted workfiles…

imccarter:
… in the dashboard's recent workfiles module. Deleted workfiles were returning null, breaking the whole list with 'undefined' values in the returned array. Preferred reduce over filtering + mapping to iterate only once.


imccarter:
This fix works, but I'm looking to see if there's a way to do it a few layers deeper. When this.model.get("data") is called, some of the objects returned have a lastOpened property, but a null workfile, which is what breaks this. The fix here just bypasses those objects, but I want to see if there's a way to not have them return in the first place.

components/frontend/app/assets/stylesheets/dashboard.scss
```diff
@@ -182,10 +182,15 @@ div[data-page="Dashboard"], div[data-page="UserDashboardEdit"] {
             position: relative;
             min-height: $dashboard-module-minheight;
 
-            .clear_list {
-                width: 75px;
-                @include panel-button($height-button-widgetshell);
-                font-size: $small-font;
+            .right {
+                @include display-flex;
```
imccarter: Having to look at the recent workfiles module for long enough I got really annoyed that the buttons to configure and 'clear list' are unaligned by one pixel. It was surprisingly stubborn, but using flex was the easiest fix.

imccarter:
Glad I'm digging deeper, this fix is broken (just less broken) and things need to be handled on the rails side.

components/core/app/models/open_workfile_event.rb
```diff
@@ -14,9 +14,11 @@ def delete_old_events_for_user
     number_of_events = OpenWorkfileEvent.where(:user_id => self.user_id).count
 
     if number_of_events > EVENT_LIMIT
-      oldest_event_id = OpenWorkfileEvent.where('user_id = ?', user_id).order('created_at DESC').limit(EVENT_LIMIT).last.id
+      oldest_event_id = OpenWorkfileEvent.where('user_id = ?', user_id).order('created_at DESC').collect{
+        |owe| owe.destroy if !owe.workfile.present?
+      }.uniq.limit(EVENT_LIMIT).last.id
```
imccarter: The .uniq here is the fix for IBX-2445

kevinmtrowbridge:
Ivan -- I owe you a review here.  Thanks for digging into the history and figuring out the real solution.

PR #345, opened 2016-05-24
DEV-15025 workspace member role dropdown styling

imccarter:
Add styling consistent with the rest of chorus for the member role dropdowns in the add/edit workspace members dialog.


jmlewis:
LGTM

PR #346, opened 2016-05-31
Visualization components cleanup

kevinmtrowbridge:
Late last year I helped Curran to create several components, when we were under time pressure to release his `Chiasm` work.  This work never made it into Chorus, but the components remained.  This area has settled down now, and it's confusing and complicated to the team to have these "tactically created" components present.

This is tracked here:

IBX-4013 Clean up visualization components
https://alpine.atlassian.net/browse/IBX-4013

This PR:
* moves the Frontend pieces of the Visualization back into the `Frontend` component
* moves the Api pieces of the Visualization back into the `Api` component
* it leaves the Core pieces (the database structures) in a new component called `Visualization`

So to summarize, this juggles some stuff around, removes VisLegacy & VisChiasm and creates a new component called Visualization.

My thinking about what an appropriate usage of the "Rails engine" style of code modularity is, has gotten more sophisticated.  

(I don't think it's the correct approach for the Backbone codebase -- we will have to come up with another approach for that.)  

I think, it's best to imagine the Api as sitting on top of a platform of Ruby functionality delivered by a collection of Rails engines.  So that's why I think it makes sense to move pieces back into Frontend and Api, while keeping the Core pieces separate.  

I hope to split the Core functionality into more pieces: Authorization, Authentication, Search, Licensing, Visualization, HDFS Database modeler, SQL Database modeler, Events stream (these are a few of the ideas) -- but I think, instead of splitting the Api into multiple pieces, we should either, just build a new one, or focus on making the existing one as terse, performant, and well documented, as possible.


kevinmtrowbridge:
TODO: Just grep through the whole codebase looking for 'chiasm' -- I have spotted a few more instances of it lying around lately.

PR #350, opened 2016-06-03
DEV-14973 workspace stage search

imccarter:
Allows the search on the workspace index page to search against the workspace stages in addition to names and owners. Tricky as the only references to the stages in the db are the generic names; "stage1", "stage2", etc.

jasmine tests still need work

@kevinmtrowbridge - I'll add comments on the possibly refactor-able parts.


components/frontend/app/assets/javascripts/collections/workspace_set.js
```diff
@@ -38,6 +54,20 @@ chorus.collections.WorkspaceSet = chorus.collections.LastFetchWins.include(
 
         params.namePattern = this.attributes.namePattern;
 
+        params.stagePattern = this.stageSearchFilter(this.attributes.namePattern);
+
         return params;
+    },
+
+    // allows for searching of workspaces by stage via their user-facing names, instead of the values in the db
+    stageSearchFilter: function(pattern) {
```
imccarter: This function essentially maps the user's search query to the stage db values, so typing "de" will return "stage1" (DEfine), "stage3" (moDEl), "stage4" (DEploy).

components/frontend/app/assets/javascripts/collections/workspace_set.js
```diff
@@ -38,6 +54,20 @@ chorus.collections.WorkspaceSet = chorus.collections.LastFetchWins.include(
...
+        var stageMatches = [];
+        _.each(this.stageNames(), _.bind(function(stage) {
+            if (pattern && stage.toLowerCase().includes(pattern.toLowerCase())) {
+                stageMatches.push(this.stageMap()[stage]);
+            }
+        }, this));
+
+        return JSON.stringify(stageMatches); // parsed in workspaces controller
```
imccarter: This JSONifying of the results was a compromise; passing just an array wasn't working, and in the spirit of separating our concerns, I opted to not convert it into the string used in the SQL query on the rails side.

components/api/app/controllers/api/workspaces_controller.rb
```diff
@@ -62,8 +62,26 @@ def index
       else
         @workspaces = workspaces.includes(succinct ? [:owner] : Workspace.eager_load_associations).order("lower(workspaces.name) ASC, workspaces.id")
 
+        stageMatches = params[:stage_pattern].present? ? JSON.parse(params[:stage_pattern]) : []
```
imccarter: This line deserves some explanation; our three possible options for the stageMatches variable are:
1. no search query = empty array
2. there IS a search query, but it doesn't match any stage-name = empty array
3. there IS a search query, and it matches one or more stage-names = array with the matching stages

components/api/app/controllers/api/workspaces_controller.rb
```diff
@@ -62,8 +62,26 @@ def index
...
+        stageMatches = stageMatches.length > 0 ? stageMatches.join("|") : "a^" # regex that will never match anything
```
imccarter: This line turns stageMatches into what we need for the SQL query. If it's an empty array, it becomes a regex that will match nothing. Otherwise, we join the stage db-names together and search against those. The joining by "|" is formatting for SQL's `SIMILAR TO` operator.

PR #351, opened 2016-06-07
Workspace stage refactor: move from string constant, into a table with an association

kevinmtrowbridge:
Move the WorkspaceStages out of a string constant in the Workspace table, and into a separate WorkspaceStage table with an association.

This simplifies:
  - searching & ordering by workspace stage
  - possible per-customer customization of the stage names
  - the localization of the stage names

-----

Hi @imccarter, @jmlewis, @mmilano: Josh spoke with me on Friday about the hoops Ivan is having to jump through in order to creating "Searching by workspace stage".

I also wasn't really all that happy with the "stage1", "stage2" keys we ended up storing in the Chorus database over the course of Chorus 6.0.

This PR is not complete, it is just to show my work so far on this theme, and get feedback.  Here's what I have so far:

1) Create a new database table called "WorkspaceStage" ... it only has 5 rows, one for each of the current Workspace Stages.  It uses https://github.com/swanandp/acts_as_list to order the stages.  The "name" column is named "name_en" -- the idea is that, we can store the localized stage names in the database.  If we did want to allow the customers to specify their own stage names, then we should store those in the database -- and further on, would we support localized, customized stage names?  The database seems like the answer to these questions.

2) Migrate the existing stages to the new setup.

3) Modify the Api ... the Workspace Stages will now be passed down in the `/api/workspaces` call like so:

```
...
has_milestones: false,
workspace_stage: {
  default: true,
  id: 1,
  name_en: "define",
  position: 0,
  complete_json: true
},
archiver: null,
...
```

You can scope or search the returned workspaces like so: `http://localhost:3000/api/workspaces.json?workspace_stage_id=2`

Additionally, you can pass an array of Workspace stage IDs, in this form: `http://localhost:3000/api/workspaces.json?workspace_stage_id=2,3`

You can get a list of all the workspace stages along with their names & IDs like so: `http://localhost:3000/api/workspace_stages.json`

This way @imccarter can do the mapping from search queries to workspace stages all on the client, and pass up any workspace stage that he needs, as an ID integer, not as a string.  

4) This PR also creates a new Admin area called "Workspace Stage naming":

alpine chorus

It allows you to edit the names of the stages:

alpine chorus-1

-----

As I mentioned, this is not ready for merging in, but please comment and let me know if this approach is preferable to that @imccarter is being forced to take in #350.  I'm sure there's plenty of time in the 6.1 timeline to fit this in as a minor feature.


jmlewis:
Awesome! Yeah it would be interesting to chat a bit more about the localization stuff and how that relates to search. I can see a few different approaches. Another thing to think about would be having the client hydrate its list of stages off the Chorus API rather than having them hard coded. Is there any point where the client gets global configuration on login?

mmilano:
the browser client gets the locale in the initial frontend
definition/construction
chorus.js.erb:

self.locale = 'en';

not clear to me where the backend gets locale.

- M.

PR #355, opened 2016-06-14
Data visibility: changes to ActiveRecord associations

kevinmtrowbridge:
This PR makes the old system of associating datasets with workspaces considerably more sophisticated.  It adds two new models WorkspaceDataSource & HdfsDataSourceConnection.

There is a "reference implementation" of the ideas contained, here: https://github.com/kevinmtrowbridge/data-vis-example (where you can see the system in isolation).

-----

There are 3 core models: "Workspaces," which have many "DataSources" and "Datasets." A "DataSource" is roughly a database server (think of Postgres). A dataset is roughly a database table inside of Postgres.

The point of the app is to allow access to the datasource "authorization" credentials to the datasources in various ways. Another goal is for specific tables inside the database to be "pinned" or highlighted within the workspace.

There are three ways that a data source can be accessible to a workspace: 

1. The DataSource is public and accessible to all workspaces. 
2. The Workspace has a corresponding DataSourceAccount associated: it has both an associated WorkspaceDataSource and a DataSourceAccount. 
3. The User has a corresponding DataSourceAccount associated: the workspace has an associated WorkspaceDataSource, but the user has the DataSourceAccount.

There is one way in which a data source can be present within a workspace, but inaccessible: 

1. The Workspace has a WorkspaceDataSource, but no DataSourceAccount can be found.

There are two types of DataSources: HdfsDataSources and regular DataSources. HdfsDataSources do not have a DataSourceAccount (username / password) for authorization, instead they have a HdfsDataSourceConnection. For brevity, this discussion uses DataSourceAccount as the example. HdfsDataSourceConnections behave similarly.

-----

models_complete

-----

This is an enumeration of all the "data visibility" possibilities for the new Workspace data tab:

janky_but_helpful_ui


components/core/db/migrate/20160620232020_create_workspace_datasets.rb
```diff
@@ -0,0 +1,10 @@
+class CreateWorkspaceDatasets < ActiveRecord::Migration
```
kevinmtrowbridge: This is the same as AssociatedDataset and should be removed, and the other renamed if that is desired.

kevinmtrowbridge:
Closing this as it is superseded by https://github.com/Chorus/chorus/pull/360

PR #356, opened 2016-06-14
Data visibility: new API routes

kevinmtrowbridge:
Ok team,

I'm at a place now where, I've got API calls for the basic stuff, for what's necessary for the Frontend.  So, on Monday let's sync up to build out the Frontend on top of this branch.

This branch has all the basic API calls we need to get started, and I created the API documentation as well.  To compile your own version of the API docs, do this:

```
git fetch origin data-visibility-6.1-core
git checkout data-visibility-6.1-core
cd component/api
./generate-api-docs.sh
--> your dev server has to be running and then, visit http://localhost:3000/api
```

The new sections look like this:

api documentation

-----

Let's imagine a walk through what the Frontend will need to do, and I will explain how these API calls can be used to accomplish it.

## API route: "Return all data sources available to the workspace, along with an explanation"
`GET /workspaces/1000008/data_sources`

First step: a Workspace is created, no Data Sources are associated with it, but when a user clicks on the Data Sources tab, the public Data Sources should be there.

Both DataSources and HdfsDataSources now have a `public` boolean attribute.  All Data Sources will be marked as "public" by default, during the migration (per @mmilano's instructions).  This attribute is serialized into the API via the DataSourcePresenter and HdfsDataSourcePresenter.

When you click on the Workspace > Data Sources tab, you'll need a list that contains a) all public data sources, and b) all data sources associated with the workspace.  The JSON includes these two keys to explain how the user / workspace will be authorized (which credentials will be passed):

    "accessible": true,
    "explain_accessibility": "user_account", 

"explain_accessibility" can be: workspace_account, user_account, shared_user_account, no_account  This is a nicety although, I suspect this might be useful as we build out the Frontend.

## API route: "Associate a data source with a workspace"
`POST /workspaces/1000008/data_sources` additional parameter: "workspace_id"

A data admin wants to associate a data source with a workspace.  This will be invoked from the Data Sources tab by a data admin.  There's also the corresponding: "Unassociate a data source from a workspace"

## API route: "Workspace data source accounts, Create workspace credentials"

You post the workspace_id, the data_source_id, the db_username, and the db_password to a URL of the form `/workspaces/1000008/data_source_account` -- this creates a new DataSouceAccount which is associated with the workspace.  When present, it will be used for connecting to the data source.  If missing, the system falls back first to individual user data source accounts, then to a shared data source account, and finally it may not be able to authenticate at all.  

These are the same situations expressed in ...

> "explain_accessibility" can be: workspace_account, user_account, shared_user_account, no_account 

There's also: update workspace credentials, and remove workspace credentials.

-----

I think with this system, we can get started on the Frontend.  We can build out the Workspace's Data Source tab.  We can build out the Data Admin functionality: associating Data Sources with workspaces and creating custom credentials for them.

## What's not here yet?

* There's no HDFS Data Source Connections yet -- this is going to take additional work, because the existing HDFS Data Source Connection is embedded in the HDFS Data Sources table ... so basically, before we can allow there to be multiple ones, I first have to extract the singular HDFS Data Source Connection, so it's stored separately and can be swapped out.

* Additionally we need to allow the Workspace specific Data Source Accounts to be passed through to at least the following APIs: Schema, Database, Dataset, HdfsDataSet -- please comment to add any additional ones, I haven't really thought this entirely through yet.


components/api/app/presenters/api/dataset_presenter.rb
```diff
@@ -73,6 +73,7 @@ def credentials_hash
 
     def associated_workspaces_hash
       return {:associated_workspaces => []} if rendering_activities?
+      # workspaces = model.workspaces.map do |workspace|
```
sxie2016: commented out code?

components/api/app/presenters/api/hdfs_data_source_presenter.rb
```diff
@@ -8,26 +8,42 @@ def to_hash
...
+                      :is_hdfs_hive => hdsc.is_hdfs_hive,
+                      :job_tracker_port => hdsc.job_tracker_port,
+                      :port => hdsc.port,
+                      :group_list => hdsc.group_list,
+                      :hive_metastore_location => hdsc.hive_metastore_location,
+                      :host => hdsc.host,
+                      :job_tracker_host => hdsc.job_tracker_host,
+                      :username => hdsc.username,
+                      :connection_parameters => hdsc.connection_parameters
+                      # :connection_parameters => model.connection_parameters_including_hive
```
sxie2016: same

kevinmtrowbridge:
Closing this as it is superseded by https://github.com/Chorus/chorus/pull/360

PR #359, opened 2016-06-16
DEV-14973 Get search by workspace stage functional with new stage models

imccarter:
Opening this PR for discussion.

The translation file is still used in a few spots; wherever the description is displayed (as that's not stored in the db), and even the names of the stages in a few spots (proper capitalization is an easily fixed, but persistent issue). See in-line comments for specific issues.


components/frontend/app/assets/javascripts/models/workspace.js
```diff
@@ -59,39 +59,49 @@ chorus.models.Workspace = chorus.models.Base.extend({
...
+        var stages = chorus.workspaceStages.models;
+        var stageMap = {};
+        _.each(stages, function(stage) {
+            stageMap[stage.id] = stage.get('nameEn');
+        });
+        return stageMap;
     },
 
-
+    // TODO: Cleanup a lot of these functions on decision of where stage name and description is stored
```
imccarter: The following couple of functions are modified so as to work without changing the translation file, using the id for the stage as reference.

components/api/app/controllers/api/workspaces_controller.rb
```diff
@@ -62,21 +62,19 @@ def index
...
 
-        if params[:workspace_stage_id].present?
-          @workspaces = @workspaces.where(:workspace_stage_id => params[:workspace_stage_id])
-        end
-
         # workspace index search filter:
         if params[:name_pattern]
           sqlQuery = <<-SQL
             workspaces.name ILIKE ? OR
+            strpos(?, workspaces.workspace_stage_id::text) > 0  OR
```
imccarter: Including the search against the workspace stage in the SQL query is a necessary evil. Rails 4.xx allows you to chain `where`s, but it does a union join, so we'd only get the workspaces where a stage AND some other attribute matched the search query. Using chained `where`s as an OR join is apparently coming in Rails 5.

components/frontend/app/assets/javascripts/presenters/activity_presenters.js
```diff
@@ -777,7 +777,7 @@
         },
 
         stage: function(self) {
-            return t('workspace.stage.' + self.model.get('stage') + '.label');
+            return t('workspace.stage.stage' + self.model.get('stage').id + '.label');
```
imccarter: One of the spots where the translation file is still used to assure proper capitalization

components/core/app/models/workspace.rb
```diff
@@ -216,6 +216,8 @@ def datasets(current_user, options = {})
 
   def self.create_for_user(user, params)
     workspace = user.owned_workspaces.build(params)
+    workspace.workspace_stage = workspace.stage # adds the (default) stage to the workspace
```
imccarter: I'm curious if it's possible to set the default stage via the params, instead of having to do it explicitly here.

components/frontend/app/assets/javascripts/chorus.js.erb
```diff
@@ -38,7 +38,10 @@ window.Chorus = function chorus$Global() {
 
         self.startHistory();
         self.updateCachebuster();
-        
+
+        self.workspaceStages = new chorus.collections.WorkspaceStages(); // namespace stages
```
imccarter: this allows the front-end stages collection to be accesible from anywhere as `chorus.workspaceStages`. `workspaceStages` will return an array of objects for all of the stages as well, but that isn't used anywhere in the code beyond creating `chorus.workspaceStages`.

components/frontend/app/assets/javascripts/dialogs/workspaces/edit_workspace_stage_dialog.js
```diff
@@ -43,30 +42,32 @@ chorus.dialogs.EditWorkspaceStage = chorus.dialogs.Base.include(
...
             // initial description for the current stage
             stageDescriptionKey: this.model.stageDescriptionLongKey(),
-            options: _.map(this.workspaceStages, function(val) {
+            options: _.map(chorus.workspaceStages.models, function(val) {
                 return {
-                    value: val,
-                    nameKey: "workspace.stage." + val + ".label",
-                    selected: this.model.get("stage") === val
+                    value: val.get("nameEn"),
+                    nameKey: "workspace.stage.stage" + val.id + ".label",
```
imccarter: Translation file used for stage name here...

mmilano: i haven't searched on this, but really - ruby must have a library of capitalization routines, including title case...mustn't it?

mmilano: https://stackoverflow.com/questions/1791639/converting-upper-case-string-into-title-case-using-ruby

components/frontend/app/assets/javascripts/presenters/activity_presenters.js
mmilano: opinionatedly, it feels unclean to still rely on the translation file because of the distance between the backend code and the strings file.

mmilano:
the title says search - but is this applying to all search, or only what is really 'filter' on the index pages of workspace?

(Search -start from no results, builds a set based on criteria that match/don't match.
Filtering - start with an existing list, and remove items based on criteria that match/don't match.)

imccarter:
Yes, this is just for the filter aspect of the workspace list. I've been using both terms somewhat haphazardly, but this isn't touching the global chorus search. 

Filtering is being handled on the back-end; that clunky constructed SQL query in the workspace controller is where the magic happens. As I noted, some of the new features in Rails 5 will obviate the clunkiness, but for us, it works, and seems pretty quick too.

components/frontend/app/assets/javascripts/dialogs/workspaces/edit_workspace_stage_dialog.js
imccarter: Ruby has a built in `capitalize`, but what I'm more concerned about is standardizing where it gets capitalized. I think we want an all lower-case value to simplify the filtering, but I'd like to avoid having to remember to capitalize it everywhere/anyhwere it's getting displayed.

imccarter:
This is pretty finalized. The last point of concern I have is migrating old stages. One of the migrations in @kevinmtrowbridge's PR (#351) handles going from stages as string-constants to objects, but I wouldn't be surprised if some more tinkering is necessary to get that one to work with the further migrations I've added here.

components/frontend/app/assets/javascripts/dialogs/workspaces/edit_workspace_stage_dialog.js
```diff
@@ -16,25 +16,22 @@ chorus.dialogs.EditWorkspaceStage = chorus.dialogs.Base.include(
...
     },
 
-    // when select list is changed, change the description (long-form) text displayed
+    // when select list is changed; change the description (long) text displayed, and (de)activate the "Update" button
     changeStage: function(e) {
-        var stageSelected = this.$("select[name='workspaceStage'] option:selected").attr("value");
-        var stageDescriptionLongKey = this.model.stageDescriptionLongKey(stageSelected);
-
-        this.$(".stage_description .message").text(t(stageDescriptionLongKey));
+        var stageNameSelected = this.$("select[name='workspaceStage'] option:selected").val();
```
jmlewis: Is there a way to get the index of the selected element from the dropdown instead of having to go through the name?

components/frontend/app/assets/javascripts/models/workspace.js
```diff
@@ -54,48 +54,10 @@ chorus.models.Workspace = chorus.models.Base.extend({
         return t(summaryStatus);
     },
 
-// *****
```
jmlewis: Such epic code destruction :D

components/frontend/app/assets/translations/messages_en.properties
```diff
@@ -2564,24 +2564,24 @@ workspace.project.progress.label=Progress:
...
-workspace.stage.stage2.description_short=Transform and cleanse the data to prepare for modeling.
-workspace.stage.stage3.description_long=Create, evaluate, and optimize machine learning models.
-workspace.stage.stage3.description_short=Create and iterate machine learning models.
-workspace.stage.stage4.description_long=Deploy models in batch scoring jobs, real time scoring APIs, and Touchpoints. Integrate the deployed models into business processes.
-workspace.stage.stage4.description_short=Deploy models and integrate them into the business processes.
-workspace.stage.stage5.description_long=Monitor and maintain the operational model. Perform minor optimizations. Measure business impact based on new behavior.
-workspace.stage.stage5.description_short=Monitor, maintain, and measure the models.
-
-workspace.stage.undefined.label=Modeling
+#workspace.stage.stage1.label=Define
```
jmlewis: Should we just delete this instead of commenting?

imccarter: We should, yes. I left it in on the off chance we need if there are some migration issues. Figure we'll take it out after/while we merge.

PR #361, opened 2016-06-27
HdfsDataSource has_many HdfsDataSourceConnections

kevinmtrowbridge:
Chorus has long had the ability to support many different DataSourceAccounts (for traditional SQL based DataSource).  However, it has never had the equivalent ability to have many Hdfs "accounts" or, a better name is `HdfsDataSourceConnectionParameterSets` or (because that is too long) `HdfsDataSourceConnections` ...

Per @jmlewis and @LawrenceSpracklen, one of the clients we are building the Data Visibility feature for, expressly wants to use the feature to provide different HdfsDataSourceConnections per workspace.  In order to make this work, we first have to build out the ability to have different HdfsDataSourceConnections, at all.

So, here is approach -- fundamental changes to allow for the creation of a HdfsDataSourceConnection model.

`HdfsDataSource has_many HdfsDataSourceConnections` ... in the same way that:
`DataSource has_many DataSourceAccounts`

Sharing this with @AndrewLngdn as he's going to help me deep dive into this.  This is one of the big remaining challenges for the Data Visibility feature.

 #data-visibility-6.1


components/core/app/models/hdfs_data_source.rb
```diff
@@ -1,21 +1,75 @@
...
+  has_many :hdfs_data_source_connections, :dependent => :destroy
+
+  def self.new_accessible_to(options)
+    workspace = options[:workspace]
+
+    ids = []
+    if workspace
+      ids = HdfsDataSourceConnection.where('workspace_id = ?', workspace).pluck(:hdfs_data_source_id)
+      # elsif user
+      #   ids = HdfsDataSourceConnection.where(user_id = ?', workspace, user).pluck(&:hdfs_data_source_id)
```
kevinmtrowbridge: This feature, would make it easi(er) to allow individual user accounts, and I mentioned this to @jmlewis -- however, he said this is not something we have to build now.

components/core/app/models/hdfs_data_source.rb
```diff
@@ -1,21 +1,75 @@
...
+      return hdfs_data_source_connections.where('workspace_id = ?', workspace.id).first
+    end
+
+    nil
+  end
+
+  def merge_connection_parameters_for(options)
+    cp = hdfs_data_source_connections_for(options)
+    attributes.merge cp.attributes
+  end
```
kevinmtrowbridge: As I discussed with @jleealpine: the APIs would handle this whole HdfsDataSourceConnection thing transparently.  That is, there would not be a separate API for `/api/workspaces/<id>/hdfs_data_sources/<id>/hdfs_data_source_connections` -- we would just, more or less, merge the HdfsDataSourceConnections into the hash representing HdfsDataSources.  This allows for a cleaner migration path for Alpine (Core) and it's arguably the right way to do it.

components/core/app/models/hdfs_data_source.rb
```diff
@@ -1,21 +1,75 @@
...
+    workspace = options[:workspace]
+
+    if public
+      return 'accessible_because_public'
+    elsif workspace && hdfs_data_source_connections.where(:workspace => workspace).any?
+      return 'accessible_because_public'
+    end
+
+    "not not_accessible"
+  end
```
kevinmtrowbridge: This branch is totally rough, I see this method is looking pretty buggy.

components/core/app/models/hdfs_data_source.rb
```diff
@@ -1,21 +1,75 @@
...
   belongs_to :owner, :class_name => 'User'
   has_many :activities, :as => :entity
   has_many :events, :through => :activities
   has_many :hdfs_entries
   has_many :workfile_execution_locations, -> { where :execution_location_type => 'HdfsDataSource' }, :foreign_key => :execution_location_id, :dependent => :destroy
-  validates_presence_of :name, :host
-  validates_presence_of :port, :unless => :high_availability?
+
+  validates_presence_of :name
+  accepts_nested_attributes_for :hdfs_data_source_connections
```
kevinmtrowbridge: As I was saying "we would just, more or less, merge the HdfsDataSourceConnections into the hash representing HdfsDataSources" ...

components/core/spec/support/fixture_builder.rb
```diff
@@ -146,15 +146,17 @@ def fixtures_dir(path = '')
...
     fbuilder.name :searchable2, HdfsEntry.create!({:path => "/searchquery/other_result.txt", :size => 11, :is_directory => false, :modified_at => "2010-10-21 22:00:00", :content_count => 22, :hdfs_data_source => hdfs_data_source}, :without_protection => true)
 
-    hdfs_hive_data_source = FactoryGirl.create(:hdfs_data_source, :name => 'hdfs_hive', :is_hdfs_hive => true,
-                                               :hive_metastore_location => "thrift://cdh5cm.alpinenow.local:9083", :owner => admin)
+    hdfs_hive_data_source_connection_parameters = FactoryGirl.create(:hdfs_data_source_connection, :is_hdfs_hive => true,
+                                                                     :hive_metastore_location => "thrift://cdh5cm.alpinenow.local:9083")
+    hdfs_hive_data_source = FactoryGirl.create(:hdfs_data_source, :name => 'hdfs_hive', :owner => admin)
+    hdfs_hive_data_source.hdfs_data_source_connections << hdfs_hive_data_source_connection_parameters
```
kevinmtrowbridge: @AndrewLngdn: this sort of thing will need to be repeated ~50x times to get the tests passing.  When I saw that, that's when I decided to back off of merging this directly into `data-visibility-6.1-core` and, do it separately / get help.

Here see attached a list of failing tests in core which came up when running the core specs with the Core component in roughly the shape of this branch:

[failing_tests.txt](https://github.com/Chorus/chorus/files/335975/failing_tests.txt)

AndrewLngdn:
@kevinmtrowbridge thanks for this! It should be helpful. I think I can get away with the same functionality with less work. I'm thinking about making the hdfs connection parameters more like the WorkspaceDataSourceAccount. I'm going to leave the original connection parameters alone, so that everything stays as it is. If you are in a context where there could be cParams attached to the workspace, it will look for those. Otherwise, it will use the 'default' parameters as we discussed a few weeks ago in one of our first meetings

PR #362, opened 2016-06-27
Data visibility 6.1: data admin role

AndrewLngdn:
@kevinmtrowbridge here are the changes needed to support the Data Admin role


kevinmtrowbridge:
Awesome! Thanks, I will review this and we can get 'er merged in.

mmilano:
@AndrewLngdn the inclusion of the data admin role in the profile page looks good.
however, the business logic of the pages needs to be altered so that any of the admin roles are exclusive from the "business user" role. 

right now, selecting either of the admin roles doesnt disable the "business user" role (except in one case), and selecting "business user" doesnt disable the 2 admin roles.

mmilano:
@AndrewLngdn also, this is probably the place where the " Administrator " role name should be changed to "Application Administrator" (dev-15109)

mmilano:
just created DEV-15227 and DEV-15228 to capture the admin/business user exclusivity.

mmilano:
@AndrewLngdn another ui issue: when creating a new user, the list of admin roles should include "none" just like the edit user page does. 
screen shot 2016-06-30 at 10 55 20 am

AndrewLngdn:
@mmilano about to push the corrections to this branch. Thanks for looking over this, it seems obvious in hindsight that a data admin cannot be a business user, but I was not sure earlier.

components/core/db/migrate/20160627203926_set_data_admin_role_on_users.rb
```diff
@@ -0,0 +1,8 @@
+class SetDataAdminRoleOnUsers < ActiveRecord::Migration
+  def change
+    User.where(admin: false).each do |user|
+      user.add_role("data_admin")
+      user.save(:validate => false)
+    end
```
kevinmtrowbridge: I think -- this may be more of a development aid -- but this migration should not be present when we release?  All existing users who are not admins, should not be migrated to be "data admins."  Right?  @mmilano can you confirm that?

mmilano:
@AndrewLngdn no - for the case of migration, the idea was that all existing users (non-admin) will be given the data admin role so that the new application environment will be like the pre-data-vizibility experience as much as possible.

in other words, by default the new upgraded environment should behave as much as possible like the old world by default. 

that is why existing data sources migrate to being public, too.

the onus is on the app owners in the upgraded system to remove the data admin role from folks that should not have it, paring down the roles, instead of having to add it to roles that should have it. 

new accounts default to not being an admin, however.

components/api/app/controllers/api/hdfs/files_controller.rb
```diff
@@ -5,6 +5,8 @@ class FilesController < ApiController
     before_filter :check_source_disabled?
 
     def index
+      authorize! :show, hdfs_data_source, current_user
```
kevinmtrowbridge: @AndrewLngdn I am making progress with this.  I am approaching it as my chance to work with the Authorization code to move it closer to what we really need.  I will quickly put my thoughts in here, so you can respond.  I will develop this though -- just tell me if you think I am off base.

For this part, I think we need to change this to be more like:

`authorize! :show, hdfs_data_source, {:user => current_user, :workspace => workspace}`

components/api/app/controllers/api/users_controller.rb
```diff
@@ -41,6 +41,12 @@ def create
         user.user_type = user_params[:user_type]
       end
 
+      if user_params.key?(:roles)
+        if defined?(Authorization::Engine)
+          user.add_role("data_admin") if user_params["roles"].include?("data_admin")
```
kevinmtrowbridge: Here, I think we decided that the whole "roles" setup is overkill for the current state of the code.  I may try to fall the "roles" back to just, being a string value.  But, I think you decided here, it wasn't actually necessary.  So I see what you're saying, but we may want to actually go ahead with the simplification.

components/authorization/app/permissions/data_source_permissions.rb
```diff
@@ -17,11 +17,12 @@ def self.user_is_data_source_owner(data_source, user)
   end
 
   def self.data_source_is_shared(data_source)
-    return true if data_source.class == HdfsDataSource
+    return false if data_source.class == HdfsDataSource
     data_source.shared?
   end
 
   def self.data_source_account_exists(data_source, user)
+    return false if data_source.class == HdfsDataSource
```
kevinmtrowbridge: Ok -- I think this file needs to be rewritten.  Here's what I have so far:

```
class DataSourcePermissions
  def self.authorized?(action, data_source, authoriziable_objects)
    case action
      when :show
        return data_source_is_public(data_source) || data_source_account_exists(data_source, authoriziable_objects)
      when :update, :destroy
        return user_is_data_source_owner(data_source, user) || user.data_admin?
      when :explore_data
        return data_source_is_public(data_source) || data_source_account_exists(data_source, authoriziable_objects)
      else
        false
    end
  end

  def self.user_is_data_source_owner(data_source, user)
    data_source.owner == user
  end

  def self.data_source_is_pubic(data_source)
    return false if data_source.class == HdfsDataSource
    data_source.shared?
  end

  def self.data_source_account_exists(data_source, authoriziable_objects)
    return false if data_source.class == HdfsDataSource
    data_source.data_source_account_for(authoriziable_objects).present?
  end
end
```

kevinmtrowbridge: `data_source.data_source_account_for(authoriziable_objects).present?`

The method I added to DataSource is called `data_source_account_for` ... it accepts hash with any combination of keys named user, workspace ... and returns the DataSourceAccount with this information.

kevinmtrowbridge:
@mmilano: Ok -- thanks for the clarification regarding the desire to migrate all users to be DataAdmins initially -- looks like @AndrewLngdn was right originally!  I will leave this part as it is.

components/api/app/controllers/api/hdfs/files_controller.rb
AndrewLngdn: Agreed, we will need to give the auth system the workspace somehow in certain cases. Maybe we could pass the workspace as an option and check it in the cases where that is relevant? Let me know what you think is simpler

kevinmtrowbridge:
This is being merged in, via other means.  Closing this PR.

PR #364, opened 2016-06-29
DEV-15207 Fix tooltips in workspace member role dropdowns

imccarter:
Having multiple dropdowns in one view was causing the tooltips for each dropdown to concatenate. This adds a conditional to check if there are multiple dropdowns and apply the proper title/tooltip to each.


jmlewis:
LGTM

PR #365, opened 2016-06-29
Data visibility 6.1 -- Frontend changes

kevinmtrowbridge:
# Theory and Practice of Workspace Data Source Visibility Scoping

**Access a branch build which exhibits this functionality here: http://10.10.7.113:8080/ -- thanks @nkapinos for the AWS machine!**

**See a Kanban board showing the progress of this feature towards completion here: https://alpine.atlassian.net/secure/RapidBoard.jspa?rapidView=37&view=detail**

## Data Admin role

data_admin_role

This is new for this branch.  Normal users can no longer see the root level "Data" screens.  Only users who have an Administration Role of "Data administrator" or "Administrator" can see these screens.

data_tab_hidden

## Workspace Data Sources tab

workspace_data_sources_tab

This is new for this branch.  Workspaces now have an additional Data Sources branch.  The Data Sources that appear here are either "public" or, they have been explicitly associated with this branch.

### Associating a "Limited" Data Source with a Workspace

This is the job of the Data admin.

1) Confirm the Data Source is not already present inside a Workspace, by navigating to its Data Sources tab.  (If the data source is present, and it has not been explicitly associated, it must be a "public" data source.)

2) Go to the root level Data tab.  Select a Data Source.  In the right hand menu, click "Associate Data Source to a Workspace:

associate_data_source_to_workspace

3) Now, navigate to the Data Sources tab of the Workspace you chose in step 1.  The Data Source should be there.

**QA task: confirm that, it all makes sense regarding "limited / public" data sources and the circumstances whereby they can be associated with Workspaces.**
**QA task: confirm that, if a user is not a Data Admin, and they are not a member of a workspace where a Data Source is associated, they cannot access the Data source.**

### Un-associating a "Limited" Data Source from a Workspace

To do this:
1) Find the Data Source in the Workspaces > Data Sources tab
2) Select the Data Source.  A link "Unassociate Data Source" should appear in the right hand menu.  Click this.
3) (For now you must reload the page.  This is temporary behavior.) The data source is no longer associated.

workspace_data_source_actions

## Creating a "Workspace Data Source Account or Credential"

The current system of linking users to Data Source Accounts or Credentials, is unchanged.  It is still possible to have a "shared" Data Source Account where a single owner credential is shared with all users.  It is still possible to have unique Data Source Accounts, one for each user.

In addition to this, it is now possible to have a Data Source Account which will override all others when the Data Source is accessed through a Workspace.

data_source_account_flowchart

To create a Workspace Data Source Account or Credential:

1) Navigate to the Workspace > Data Sources tab.
2) Find a limited (non public) Data Source and select it.
3) In the right hand menu, find a link "Edit Credentials" -- click this.  A dialog appears where you can edit the Workspace specific username and password.

**QA task: the database can be configured to return different information for the same queries, depending on which account is used to run the queries.  Enabling that is the whole purpose of this feature!  :)  The task will be to setup different accounts for the same Data Source, for different workspaces, and confirm that the data displayed differs.**

## Drilling down into a Data Source within a Workspace

All of the old Data Source behavior which (is still present) on the Data tab, is also accessible when drilling down into a data source from inside of a workspace.

**QA task: ensure that all of this behavior is consistent with the use of the Workspace Specific Data Source accounts.**

# What is not done?

* The UI is still rough when drilling down into Data Sources inside of Workspaces.  The pages have not yet been adjusted to show that we are still "inside" of a workspace.

alpine chorus

* There is a security hole currently when using the search function.  Non data-admins can see Data Sources which they are not authorized to see, when searching.  We are working on this.

* The new dialogs are still rough.  We intend to work the "Associate Data Source with Workspace" process into more of a wizard, not a two step process.  @imccarter and @mmilano are working on this.


components/frontend/app/assets/javascripts/collections/database_set.js
```diff
@@ -1,8 +1,13 @@
...
-    chorus.Mixins.DataSourceCredentials.model
+    chorus.Mixins.DataSourceCredentials.model,
+    chorus.Mixins.WorkspaceScopingHelpers
 ).extend({
     constructorName: "DatabaseSet",
     model: chorus.models.Database,
-    urlTemplate: "data_sources/{{dataSourceId}}/databases",
-    showUrlTemplate: "data_sources/{{dataSourceId}}/databases"
+    urlTemplate: function() {
+        return "data_sources/{{dataSourceId}}/databases" + this.workspaceIdParamOrBlank();
```
kevinmtrowbridge: Here is an example of using `Backbone.history.getFragment` to dynamically
pass along the `workspace_id` when present.

components/frontend/app/assets/javascripts/pages/dataset/database_index_page.js
```diff
@@ -1,18 +1,43 @@
...
+                {label: this.workspace.get("name")}
+            );
+        }
+
+        array.push(
+            {label: t("breadcrumbs.data_sources"), url: "#/data_sources"},
+            {label: this.dataSource.get("name")}
+        );
+
+        return array;
```
kevinmtrowbridge: Ivan: this above is an example of how the breadcrumbs can be dynamically modified to react to whether or not the page is being displayed inside a workspace or not.

components/frontend/app/assets/javascripts/pages/dataset/database_index_page.js
```diff
@@ -1,18 +1,43 @@
...
     },
 
-    setup: function(dataSourceId) {
-        this.dataSource = new chorus.models.GpdbDataSource({id: dataSourceId});
+    setup: function() {
+        this.workspaceId = this.parseWorkspaceIdFromRoute();
```
kevinmtrowbridge: Here is how virtually any file can switch its behavior around depending on whether or not the workspaceId is present in the URL.

components/frontend/app/assets/javascripts/pages/dataset/database_index_page.js
```diff
@@ -1,18 +1,43 @@
...
+        // page can either be accessed from inside a workspace, or via the root level data tab
+        if(!this.workspaceId) {
+            this.dataSourceId = arguments[0];
+            this.dataSource = new chorus.models.GpdbDataSource({id: this.dataSourceId});
+
+        } else {
+            this.dataSourceId = arguments[0];
+            this.workspace = new chorus.models.Workspace({id: this.workspaceId});
+            this.dataSource = new chorus.models.GpdbDataSource({id: this.dataSourceId, workspace_id: this.workspaceId});
+            this.subNav = new chorus.views.SubNav({workspace: this.workspace, tab: "datasources"});
```
kevinmtrowbridge: Here, the `DatabaseIndexPage` is adding the Workspace tabs SubNav to itself, if a Workspace is present.

components/frontend/app/assets/javascripts/router.js
```diff
@@ -71,12 +72,12 @@
...
-            ["data_sources/:dataSourceId/databases", "DatabaseIndex"],
+            ["data_sources/:dataSourceId/databases(*workspaceIdParam)", "DatabaseIndex"],
             ["data_sources/:id/schemas", "OracleSchemaIndex"],
 
-            ["databases/:databaseId", "GpdbSchemaIndex"],
-            ["schemas/:schemaId", "SchemaDatasetIndex"],
-            ["datasets/:id", "DatasetShow"],
+            ["databases/:databaseId(/*workspaceIdParam)", "GpdbSchemaIndex"],
+            ["schemas/:schemaId(/*workspaceIdParam)", "SchemaDatasetIndex"],
+            ["datasets/:id(/*workspaceIdParam)", "DatasetShow"],
```
kevinmtrowbridge: The routes need to be modified to optionally respond to the "workspace_id" parameters.  The parentheses mark parts of the routes as optional.  

The URL is a little strange in that there's a trailing slash (/) before the question mark:

`/#/data_sources/1/databases/?workspace_id=1`

(Backbone is a little weird with query parameters: https://github.com/jashkenas/backbone/issues/891)

Backbone documentation: http://backbonejs.org/#Router-routes

PR #366, opened 2016-06-29
Data visibility 6.1: Core / API scoping & authorization by workspace_id

kevinmtrowbridge:
## First goal: data controllers accessible by non data-admins when scoped by a workspace_id

This is how I started out with this branch:

I went to the Data Sources tab, and clicked on a DataSource.  I saw that, it loaded all the schemas from the SchemasController, so I thought, I have to modify the SchemasController to accept a workspace_id.

I did that as quickly as possible, and I added a test to the SchemasControllerSpec.

Then I proceeded to click on a Schema in the UI, and I saw it took me to the DatabasesController, so I thought, I have to modify the DatabasesController to accept a workspace_id.

I did that, and then I modified the DatabasesControllerSpec to have a test for the case where a workspaces_id is passed in.

So, the SchemasController and the DatabasesController have the first piece of the work done.

Here’s a partial list of controllers which still need work: 
* chorus_views_controller.rb
* database_schemas_controller.rb
* datasets_controller.rb
* previews_controller.rb
* streams_controller.rb
* workfile_executions_controller.rb
* db_execute_controller.rb

## Second goal: authorization sanity check for data admins and non-data admins

If a non-data-admin clicks on a SchemasController and does not pass a workspace_id, there needs to be an authorization failure of some sort.  Or — maybe it just returns an empty array?  Not quite sure yet, suspect it will be clear in the individual cases.

## 3rd goal: complete modifications to core

We still have to figure out how these files need to change:
* csv_copier.rb
* gnip_copier.rb
* gpfdist_table_copier.rb
* multi_pg_table_copier.rb
* oracle_table_copier.rb
* table_copier.rb


components/core/app/models/data_source.rb
```diff
@@ -190,10 +190,11 @@ def connect_as_owner
     connect_with(owner_account)
   end
 
-  def connect_as(user)
-    connect_with(account_for_user!(user))
+  def connect_as(account_finder_params)
+    connect_with(data_source_account_for(account_finder_params))
   end
```
kevinmtrowbridge: This is the 1st, small, central change of this PR.  DataSource#connect_as, instead of taking a singular `user` object, it now takes a hash like this: `{:user => user, :workspace => workspace}`.

(One or none of these can be passed.  Although I guess for right now it would have to be an empty hash like `{}`, we can modify it to be `def connect_as(account_finder_params = {})` ... data_source_account_for({}) can work if the DataSource is shared.

components/core/app/models/database.rb
```diff
@@ -55,20 +57,20 @@ def find_dataset_in_schema(dataset_name, schema_name)
...
 
-  def create_schema(name, current_user)
+  def create_schema(name, account)
     new_schema = schemas.build(:name => name)
     raise ActiveRecord::RecordInvalid.new(new_schema) if new_schema.invalid?
 
-    connect_as(current_user).create_schema(name)
-    Schema.refresh(account_for_user!(current_user), self)
+    connect_with(account).create_schema(name)
+    Schema.refresh(account, self)
```
kevinmtrowbridge: In general, where possible, I have switched over to passing along the DataSourceAccount, instead of the user.  Maybe this is wrong, we could instead modify `DataSource#connect_as` to examine what it receives and react accordingly.

components/core/app/models/gpdb_table.rb
```diff
@@ -5,8 +5,8 @@ def analyze(account)
     schema.connect_with(account).analyze_table(name)
   end
 
-  def verify_in_source(user)
-      schema.verify_in_source(user) &&
-      schema.connect_as(user).table_exists?(name)
+  def verify_in_source(account)
+      schema.verify_in_source(account) &&
+      schema.connect_with(account).table_exists?(name)
   end
```
kevinmtrowbridge: This is the other "big change" in Core with this PR -- there are a bunch of these `verify_in_source` methods, which used to take `user` but, now I have modified them to take `account` -- I really think this change especially makes sense in this case.

kevinmtrowbridge:
This is done, except for some failing tests.  We can still fix the failing tests, in the feature branch.  And we need this work to meet up with Ivan's Frontend work tomorrow.  For that reason, I will merge this into the main Data Viz feature branch now.  We will work on fixing the tests to be perfect, once we have the main UI flows working.  QA can also help with finding broken things.

PR #379, opened 2016-07-22
Data visibility 6.1: explore adding authorization type

kevinmtrowbridge:
Marks made in the code to indicate areas that would need to be modified in order to add in a new attribute to DataSource, called "authorization_type" which, would replace the use of "shared" in the current "cascading" account retrieval scheme.

This is only in the application code, not in any tests.


components/core/app/models/data_source.rb
```diff
@@ -39,16 +52,13 @@ def data_source_account_for(authorizable_objs_or_ids = {})
...
-
-    if user
-      acct ||= data_source_accounts.where(:owner => user).first
+    case authorization_type
+      when "shared_for_all_workspaces"
+        acct = owner_account
+      when "different_for_each_workspace"
+        acct = data_source_accounts.where(:workspace => workspace).first
+      when "individual"
+        acct = data_source_accounts.where(:owner => user).first
```
kevinmtrowbridge: So this is the major, core change.  It no longer cascades, it merely works according to the authorization_type.

PR #380, opened 2016-07-25
Dev 15291 sample data experiment

sxie2016:
Introduced an extra system configure option during chorus installation to allow customer whether installing demo sample data or not. Still working on cleanup the code but it's ready for initial review. @haozou @nkapinos


packaging/create_workspace.py
```diff
@@ -0,0 +1,23 @@
...
+import requests
+import json
+
+session = requests.session()
+session.headers.update({"Host": "localhost:8080"})
+session.headers.update({'content-type': 'application/x-www-form-urlencoded'})
+
+url = "http://localhost:8080/sessions?session_id=NULL"
+body = {"username": "demo_user", "password": "demo_user"}
+print("Creating Chorus Session...")
```
sxie2016: would clean this up

haozou: from Log import logger

packaging/load_demo_sample_data.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
haozou: bin=`readlink "$0"`
if [ "$bin" == "" ]; then
 bin=$0
fi
bin=`dirname "$bin"`
bin=`cd "$bin"; pwd`

. "$bin"/chorus-config.sh

$PYTHON $CHORUS_HOME/current/packaging/seed.py

packaging/setup_sample_demo_data.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
haozou: bin=readlink "$0"
if [ "$bin" == "" ]; then
bin=$0
fi
bin=dirname "$bin"
bin=cd "$bin"; pwd

. "$bin"/chorus-config.sh

packaging/upload_workfiles.py
```diff
@@ -0,0 +1,88 @@
...
+logger.info("Token: {0}".format(token))
+
+
+DATA_PATH = "$CHORUS_HOME/packaging/demo_sample_data"
+
+# get workspace id by workspace name
+def get_workspace_id(workspace_name=WORKSPACE_NAME):
+  url = "{0}/workspaces?session_id={1}".format(API_URL, token)
+  payload = ''
+  workspaces = session.get(url, data=payload, verify=False)
```
nkapinos: @sxie2016 You don't need to provide a payload for REST GET requests.  You can delete line 25 payload="', and remove data=payload from line 26.

packaging/sample_data_constants.py
```diff
@@ -0,0 +1,35 @@
+# API CONSTANTS
+API_HOST = "localhost"
+API_PORT = 8080
+API_PROTOCOL = "http"
+API_URL = "{0}://{1}:{2}".format(API_PROTOCOL, API_HOST, API_PORT)
+
```
nkapinos: These constants are OK for our first iteration, but let's think about how to handle installs with HTTPS / different ports for 6.2.

sxie2016: Sure. my thought is that might be we could put it into the config or command option. so that we could easy to override it.

packaging/upload_workfiles.py
sxie2016: Thanks @nkapinos. addressed.

PR #384, opened 2016-07-26
Dev15359: Allow Alpine to read read license and credentials API when HTTPS certificate uses FQDN

nkapinos:
@kevinmtrowbridge @jleealpine @AndrewLngdn  Please review.

This fix allows a user to specify the FQDN of the Chorus host in the HTTPS certificate and open Alpine workflows.  Previously, we only allowed certificates that identified themselves as 'localhost' to access the /licenses and /datasources/id/credentials API.  This caused the requests from Alpine to fail with a 401 (reading license info), then 404 (getting database credentials).  The new behavior is to allow 'localhost' and the value set in chorus.properties 'public_url'.

If a customer wants to use a SSL certificate in production, they'll likely use the FQDN of their Chorus host, which gets passed down to the alpine and licenses controllers as 'request.host'.


nkapinos:
The changes look good.  Thanks @kevinmtrowbridge

PR #386, opened 2016-07-27
Data Vis 6.1 ui fixes

imccarter:
This fixes various issues with workspace credential creation/deletion and:

DEV-15375
DEV-15370
DEV-15408
and the fix mentioned in Josh's comment on DEV-15381

DEV-15407 too, now


components/api/app/controllers/api/workspaces/data_source_accounts_controller.rb
```diff
@@ -36,7 +36,7 @@ def update
     end
 
     def destroy
-      data_source_id = params[:account][:data_source_id]
+      data_source_id = params[:data_source_id] || params[:account][:data_source_id]
```
imccarter: This is the only change that was absolutely necessary; params[:account] was nil, so it was failing without. Kevin mentioned changing the others if this was needed.

components/frontend/app/assets/javascripts/models/data_source_account.js
```diff
@@ -7,10 +7,13 @@ chorus.models.DataSourceAccount = chorus.models.Base.extend({
         var method = options && options.method;
         var isEditingOwnAccount = this.get("userId") === chorus.session.user().id;
         var isWorkspaceAccount = !!this.get("workspaceId");
-
         if (isWorkspaceAccount) {
             this.workspaceId = this.get("workspaceId");
-            return "workspaces/" + this.workspaceId + "/data_source_account";
+            if (method === "delete") {
+                return "workspaces/" + this.workspaceId + "/data_source_account/?data_source_id={{dataSourceId}}";
```
imccarter: This is the ugly part where I hardcode a query-string param. I think a better fix may be a small API change. Thoughts, @kevinmtrowbridge or @AndrewLngdn ?

kevinmtrowbridge:
Hi @imccarter I'm sorry I can't get to this tonight!  But I will review it tomorrow first thing.

components/frontend/app/assets/javascripts/models/data_source_account.js
kevinmtrowbridge: Hmm I mean, at the least it's quite explicit ... I may prefer that actually.

kevinmtrowbridge:
This may be a good issue to piggy-back onto this branch: https://alpine.atlassian.net/browse/DEV-15407

kevinmtrowbridge:
Ok -- other than DEV-15407 I don't think I have any other comments.  This all looks good.  I don't think we need to change that query param right now, @imccarter.  That whole controller is a bit odd -- not sure we need to fix it right now.

imccarter:
Sounds good, thanks @kevinmtrowbridge! I will work on committing DEV-15407 onto this today.

PR #387, opened 2016-07-27
DEV-15308 DEV-15309 dataset association fixes

imccarter:
Small fix for dataset association options while within a workspace.


jmlewis:
LGTM

PR #389, opened 2016-07-28
Data viz 6.1 link fixes

imccarter:
This fixes various links and such:
DEV-15371
DEV-15405
DEV-15404


imccarter:
Going to try and add a fix for DEV-15414 onto this.

